Issue #396, opened 2013-04-18
Support basic HTML tags in Ticker


bamnet:
Ticker Text should support a whitelist of simple HTML tags:

``` html
<b>
<i>
<em>
<br />
```

PR #459, opened 2013-05-23
generalize preview to content types

mfrederickson:
addresses items Brian brought up in #396 
@bamnet, I'll let you merge this one when you're happy with it.

### DIFF
--- a/app/views/contents/ticker/_form_top.html.erb
+++ b/app/views/contents/ticker/_form_top.html.erb
@@ -9,7 +9,7 @@
           <a href="#tab1" data-toggle="tab"><%=t('.ticker_text')%></a>
         </li>
         <li>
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -8,7 +8,7 @@
   #TODO(bamnet): Clean this up
   match "update_owners" => "screens#update_owners"
 
-  match "ticker_preview" => "contents#ticker_preview"
+  match "preview" => "contents#preview"
```
bamnet: We can make this cleaner, I need to clean up some of our matches (we have too many).

Add the a block like the following under resources :contents

```
    collection do
      post :preview
    end
```

which will build a named route like `preview_contents_path` you can use in the view.

Issue #446, opened 2013-05-21
user membership layout suggestions


mfrederickson:
If we omit the email, change the radio buttons to dropdowns and merge the list of members and the list of leaders and distinguish them somehow, and move the update permissions button to under the list, I think it would simplify the screen quite a bit.  It would also be handy to link the users name to their user record.

![image](https://f.cloud.github.com/assets/473165/543791/7c151192-c242-11e2-8a68-a3485d3732f1.png)

PR #557, opened 2013-06-15
Issue #446 

gbprz:
Made the changes recently mentioned in Issue #446 to show drop down menus when user is logged in or labels when user is not logged in.

### DIFF
--- a/app/views/groups/_show_body.html.erb
+++ b/app/views/groups/_show_body.html.erb
@@ -56,7 +56,7 @@
                       <td>
                         <ul class="inputs-list">
                             <li> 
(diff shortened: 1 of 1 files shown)


app/views/groups/_show_body.html.erb
```diff
@@ -56,7 +56,11 @@
                       <td>
                         <ul class="inputs-list">
                             <li> 
-                              <%= select_tag "#{m.object_name}[perms][#{perm.to_s}]", options_for_select(levels.keys, m.object.perms[perm]) %>
+                              <% if can? :update, Membership.new(:group => @group, :level => Membership::LEVELS[:pending]) %>
```
bamnet: I think you can get away with not creating a new Membership object.  Maybe something like `can? :update, m`

Issue #418, opened 2013-05-02
Build generic content#index for non-feed browsing


bamnet:
It would be useful to have a generic content#index which could be used for API calls or other non-feed based browsing.  As an example, it would be useful to do something like /content?user=1 to get all the content owned by user 1 or content?screen=3 to see all the content being shown on screen 3.

PR #634, opened 2013-08-14
Content index

gbprz:
#418

First iteration seems to work well, let me know what you think

The solution seemed to be getting too long to leave it in the contents_controller, so I put it in the contents model. Was this a bad move? I can bring everything back to the contents_controller if this is not good style.

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -1,0 +1,4 @@
+load 'deploy'
+
+load 'deploy/assets' # Uncommented if you are using Rails' asset pipeline
+load 'config/deploy' # remove this line to skip loading any of the default tasks
--- a/app/assets/stylesheets/application/common/concertocons.scss.erb
+++ b/app/assets/stylesheets/application/common/concertocons.scss.erb
@@ -3,7 +3,7 @@
 // -------------------------------------------
 
 
-$concertoconsPath: "/assets/concertocons" !default;
+$concertoconsPath: font-path('concertocons') !default;
 
 @font-face {
   font-family: "concertocons";
--- a/app/assets/stylesheets/application/common/font-awesome.scss.erb
+++ b/app/assets/stylesheets/application/common/font-awesome.scss.erb
@@ -24,7 +24,7 @@
  */
 /* FONT PATH
  * -------------------------- */
-$fontAwesomePath: "/assets/fontawesome-webfont" !default;
+$fontAwesomePath: font-path("fontawesome-webfont") !default;
 
 @font-face {
   font-family: "FontAwesome";
--- a/app/views/screens/_form.html.erb
+++ b/app/views/screens/_form.html.erb
@@ -1,28 +1,28 @@
 <%= form_for(@screen) do |f| %>
   <%= render :partial => "layouts/errors", :object => @screen %>
 
-	<fieldset>
+  <fieldset>
     <legend><span><%= t('.provide_details') %></span></legend>
-		
+
 
     <div class="row-fluid">
       <div class="span4">
         <div class="clearfix">
-    			<%= f.label :name %>
-    			<div class="input">
-    				<%= f.text_field :name, :class => "input-xlarge", :autofocus => true %>
-    			</div>
-    		</div>
-    		
-    		<div class="clearfix">
-    			<%= f.label :location %>
-    			<div class="input">
-    				<%= f.text_field :location, :class => "input-xlarge" %>
-    			</div>
-    		</div>
+          <%= f.label :name %>
+          <div class="input">
+            <%= f.text_field :name, :class => "input-xlarge", :autofocus => true %>
+          </div>
+        </div>
 
         <div class="clearfix">
-          <label><%= t('.visibility') %></label>
+          <%= f.label :location %>
+          <div class="input">
+            <%= f.text_field :location, :class => "input-xlarge" %>
+          </div>
+        </div>
+
+        <div class="clearfix">
+          <%= f.label :is_public %>
           <div class="input checkbox">
             <ul class="inputs-list">
               <li><%= f.check_box :is_public %> <%= f.label :is_public %></li>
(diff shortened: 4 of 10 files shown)


app/models/content.rb
```diff
@@ -179,4 +179,33 @@ def perform_action(action_name, options)
...
+          if params[:type] && content.type != params[:type] then next end
+          filtered_contents.push(content)
+        end
+      end
+    else
+      # If filtering by user or type, we don't need to search through subscriptions, only contents
+      if params[:user] then query_conditions[:user_id] = params[:user] end
+      if params[:type] then query_conditions[:type] = params[:type] end
+      filtered_contents = Content.find(:all, :conditions => query_conditions)
+    end
```
zr2d2, line 214: I think these would be cleaner written as

``` ruby
query_conditions[:feed_id] = params[:feed] if params[:feed]
query_conditions[:screen_id] = params[:screen] if params[:screen]
if query_conditions.empty?
  # code if true
else
  # code if false
end
```

I favor the `if` statements at the end, then you can drop the then and end statements

app/views/contents/index.html.erb
```diff
@@ -0,0 +1,26 @@
+<section class="viewblock">
+  <header class="viewblock-header" id="browse-header">
+    <h1 class="default-padding"><%= t('.Concerto Content') %> </h1>
```
zr2d2: you don't necessarily need the parentheses

gbprz: Okay, this also reminds me that I need to add the translation

app/models/content.rb
gbprz, line 214: Fixed in my latest commits, thanks for the heads up. Is it good to style the if statement before the 'next' the same way? I did so for lines 196 and 197 in my most recent commits

app/controllers/contents_controller.rb
```diff
@@ -1,5 +1,5 @@
 class ContentsController < ApplicationController
-  before_filter :get_content_const, :only => [:new, :create, :preview]
+  before_filter :get_content_const, :only => [:index, :new, :create, :preview]
```
bamnet: I don't think you're using the global that get_content_const defines, are you?

config/routes.rb
```diff
@@ -165,9 +166,10 @@ def matches?(request)
   match 'content/' => 'feeds#index'
   match 'browse/' => 'feeds#index'
 
+
```
bamnet: Care to remove this extra newline?

app/models/content.rb
```diff
@@ -179,4 +179,39 @@ def perform_action(action_name, options)
   def after_add_callback(unused_submission)
   end
 
+  def self.filter_all_content(params)
+    # Filter Concerto content by user, screen, feed, and/or type
+    filtered_contents = Array.new
+
+    query_conditions = {}
+    # If filtering by screen or feed, we must search through subscriptions
```
bamnet: This isn't anywhere close to how I would have approached it (chaining where clauses) but it looks pretty neat, nice work!

config/routes.rb
mfrederickson: lol.  sounds reminiscent of our tab discussion on tuesday.

app/models/content.rb
```diff
@@ -179,4 +179,39 @@ def perform_action(action_name, options)
...
+    if params[:screen] || params[:feed]
+      query_conditions[:feed_id] = params[:feed] if params[:feed]
+      query_conditions[:screen_id] = params[:screen] if params[:screen] 
+      subs = Subscription.find(:all, :conditions => query_conditions)
+      subs.each do |sub|
+        sub.contents.each do |content|
+          # If filtering by user or type, do not add (skip) content that 
+          # does not match filter criteria
+          next if params[:user] && content.user_id.to_s != params[:user]  
+          next if params[:type] && content.type != params[:type]
```
zr2d2, line 199: nice work, I like it!

app/models/content.rb
```diff
@@ -179,4 +179,39 @@ def perform_action(action_name, options)
...
 
+  def self.filter_all_content(params)
+    # Filter Concerto content by user, screen, feed, and/or type
+    filtered_contents = Array.new
+
+    query_conditions = {}
+    # If filtering by screen or feed, we must search through subscriptions 
+    if params[:screen] || params[:feed]
+      query_conditions[:feed_id] = params[:feed] if params[:feed]
+      query_conditions[:screen_id] = params[:screen] if params[:screen]
```
zr2d2: I still might put the `query_conditions` outside the `if` statement and use `if !query_conditions.empty?`, or even switch the if/else, and not negate the `.empty?`

gbprz: Not sure if I am understanding properly, but replace:

```
if query_conditions.empty?
  filtered_contents = Content.find(:all)
else
  filtered_contents = Content.find(:all, :conditions => query_conditions) 
end 
```

with

```
filtered_contents = Content.find(:all) if query_conditions.empty? 
filtered_contents = Content.find(:all, :conditions => query_conditions) if !query_conditions.empty? 
```

config/routes.rb
gbprz: hahaha, yea ... my bad 
:octocat:

app/models/content.rb
```diff
@@ -179,4 +181,39 @@ def perform_action(action_name, options)
...
+    else
+      # If filtering by user or type, we don't need to search through subscriptions, only contents
+      query_conditions[:user_id] = params[:user] if params[:user]
+      query_conditions[:type] = params[:type] if params[:type]
+      if query_conditions.empty?
+        # No filters are specified
+        filtered_contents = Content.all
+      else
+        # User and/or type filters are specified
+        filtered_contents = Content.all query_conditions
```
zr2d2, line 212: I guess you could do that down here and save a few lines, but i'm not too concerned either way

app/models/content.rb
```diff
@@ -179,4 +181,39 @@ def perform_action(action_name, options)
...
+          filtered_contents.push(content)
+        end
+      end
+    else
+      # If filtering by user or type, we don't need to search through subscriptions, only contents
+      query_conditions[:user_id] = params[:user] if params[:user]
+      query_conditions[:type] = params[:type] if params[:type]
+      if query_conditions.empty?
+        # No filters are specified
+        filtered_contents = Content.all
```
zr2d2, line 209: .find(:all) is depreciated, so i changed it to .all

PR #683, opened 2013-08-28
handle slide transition and fieldconfig on frontend

### LINKED ISSUE #397: Allow different date/time formats in the frontend
ClientTime picks one date format, we should let the users change this to match their preferences.

### LINKED ISSUE #649: add slide transition for frontend content
@bamnet I've pretty much got this working (see branch).  However, I am setting the default transition in the position class (based on the passed in field_config)-- should I be doing this in the field class?  If so, why have an argument for the transition passed in from the call to the field construction from the position class?

There's a quirk with sliding the graphic that I am still working on (seems to disappear instead of sliding out-- but it does slide in).  I also had to change the img to have the margin and not the div.  But of course, the side effect of this is the border (specified in the position table's style column for graphic) is applied to the div instead of the img.  What do you recommend?

For the slider to work, I also had to set position:absolute on the content-- this has had no side-effects on the Fade transition.

Once I get this wrapped up, I'm hoping to work on not performing a transition if the next content is the same as the current content.

mfrederickson:
This is ready to merge once someone reviews it.

To test the transitions: add a field config entry of `transition` with one of the following values: slide, replace, _fade_ (default).  This fixes #649.

To test the datetime formatting: add a field config entry of `format` with the desired format specification.  The default is the _localized medium date and short time_ (this hasn't changed).  This fixes #397.

Format specifiers can be found here:
http://docs.closure-library.googlecode.com/git/closure_goog_i18n_datetimeformat.js.source.html#line32


public/frontend_js/content.js
```diff
@@ -31,6 +31,13 @@ concerto.frontend.Content = function(data) {
   this.id = data['id'] || null;
 
   /**
+   * The Content Type.
+   * @type {?string}
+   * @private
+   */
+  this.type_ = data['type'] || 'unknown';
```
bamnet, line 38: What's this used for?  It seems redundant to store the name of the class since we can deduce it from ContentTypeRegistry.

mfrederickson, line 38: Just for logging info like in line 74.

public/frontend_js/content.js
```diff
@@ -85,9 +92,14 @@ concerto.frontend.Content.prototype.startLoad = function() {
  * Apply the sanitized position styles to the content div.
  *
  * @param {Object} styles The styles to be applied to this content.
+ * @param {?Element} element to receive the styles - defaults to this.div_
  */
 concerto.frontend.Content.prototype.applyStyles = function(styles) {
-  goog.style.setStyle(this.div_, styles);
+  var el = this.div_;
+  if (arguments.length > 1 && arguments[1] != null) {
```
bamnet: You should explicitly add element to the args for this function instead of grabbing it from styles.  Did lint.sh complain here?

public/frontend_js/contents/graphic.js
```diff
@@ -97,8 +97,9 @@ concerto.frontend.Content.Graphic.prototype.loaderFinish_ = function(e) {
  * @override
  */
 concerto.frontend.Content.Graphic.prototype.applyStyles = function(styles) {
-  concerto.frontend.Content.Graphic.superClass_.applyStyles.call(this, styles);
-  goog.style.setBorderBoxSize(this.div_,
+  concerto.frontend.Content.Graphic.superClass_.applyStyles.call(this, styles, goog.dom.getFirstElementChild(this.div_));
```
bamnet: I have a hunch that the linter (which we probably should have documented somewhere) will complain about the line length here.

public/frontend_js/content.js
mfrederickson: I was trying to make it an optional argument, is that not how it is done?  I didn't think it was pulling it from styles.  If lint did complain, I didn't see it.  I'll check again.

mfrederickson: Does lint run automatically during the compile process?

bamnet: Negative, they're two separate steps (for no particular reason).

Unlike Java, JavaScript allows you to call a method without all of its args.

foo = function(a, b) can be called with just foo("bar").

public/frontend_js/field.js
```diff
@@ -27,9 +26,17 @@ goog.require('goog.structs.Queue');
  * @extends {goog.events.EventTarget}
  */
 concerto.frontend.Field = function(position, id, name, content_path,
-                                   opt_transition) {
+                                   opt_transition, config) {
   goog.events.EventTarget.call(this);
 
+  if(goog.isDefAndNotNull(config)) {
```
bamnet: This looks like debug code.  Maybe you can surround it in `if (goog.DEBUG) {...` which will automatically remove it when it's compiled in non-debug mode (me thinks).

public/frontend_js/field.js
```diff
@@ -88,7 +95,14 @@ concerto.frontend.Field = function(position, id, name, content_path,
...
    */
-  this.transition_ = opt_transition || concerto.frontend.Transition.Fade;
+  this.transition_ = opt_transition;
+
+  /**
+   * Configuration properties for the field
+   * @type {!Object}
+   * @private
+   */
+   this.config_ = (goog.isDefAndNotNull(config) ? config : null);
```
bamnet: No need for the ternary, just set it to config.

public/frontend_js/content.js
mfrederickson, line 38: I'm not sure how to do that.

bamnet, line 38: Feel free to leave it in and open an issue for me to "Deduce content type name from ContentTypeRegistry for debugging".

Off the cuff, I might add a method to Content which iterates over ContentTypeRegistry (which is a dict of ['string'] = Class) and checks `if (this instanceof klass) { return key; }` where `klass` is the value from ContentTypeRegistry.

public/frontend_js/position.js
```diff
@@ -109,12 +112,31 @@ concerto.frontend.Position.prototype.load = function(data) {
 
   this.draw();
 
+  transition = concerto.frontend.Transition.Fade;
+  config = null;
+  if (goog.isDefAndNotNull(data['field']['config'])) {
+    config = data['field']['config'];
+    if (goog.isDefAndNotNull(config.transition)) {
```
bamnet: You should use config['transition'].  If you try and access it as a property it might get obfuscated away.

public/frontend_js/position.js
```diff
@@ -109,12 +112,31 @@ concerto.frontend.Position.prototype.load = function(data) {
 
   this.draw();
 
+  transition = concerto.frontend.Transition.Fade;
```
bamnet: If this is the first time you're using transition you should declare it with var as well. Same applies to config below.

Issue #230, opened 2012-10-21
Support PDFs and other interesting file formats for content


mikldt:
Add support (in the form of a content-type plugin or some sort of sub-content-type plugin) to facilitate conversion "from strange format to better format."

Inputs we may want to accept would be:
1. PDF
2. Bitmaps
3. Tiff
4. Raw
5. Webp (actually we’ll do this natively)

This could take a lot of forms, so let's discuss.

PR #702, opened 2013-09-05
convert files a.k.a. process graphics 

mfrederickson:
### Requirements

This uses the _docsplit_ gem (which I've added to the Gemfile), which requires the following apt-get packages (on Ubuntu, they are:)
- graphicsmagick
- poppler-utils and poppler-data

It's more functional if you have [additional software loaded](http://documentcloud.github.io/docsplit/).
### Limitations
- It only converts the first page of any document.  (This is just the first phase.  Phase two will process multiple pages via a delayed job as child content.)
- It uses /tmp for temporary files (imagine that!)
- No tests have been written (I'm not sure how we want to handle those)

This addresses ticket #230 [and #4].

### DIFF
--- a/lib/concerto/content_converter.rb
+++ b/lib/concerto/content_converter.rb
@@ -1,16 +1,9 @@
 module Concerto
 
+  # Class to convert various mime types for content media
   class ContentConverter
-    # To add another converter, add the array of what it handles here, and then in the handles? method
-    # check that array also.  In the convert method call your converter.
-    NULL_TYPES = [
-      "image/gif",
-      "image/jpeg",
-      "image/pjpeg",
-      "image/png",
-      "image/svg+xml",
-      "image/tiff"
-    ]
+    # To add another converter, add the array of what it handles here, and then in 
+    # the supported_types method also.  Then in the convert method call your converter.
     DOCSPLIT_TYPES = [
       "application/msword",
       "application/pdf",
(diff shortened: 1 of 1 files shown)


app/models/graphic.rb
```diff
@@ -1,9 +1,8 @@
 class GraphicValidator < ActiveModel::Validator
   # Validator for Media associated with a Graphic.
   def validate(record)
-    graphic_types = ["image/gif", "image/jpeg", "image/pjpeg", "image/png", "image/svg+xml", "image/tiff"]
-    if !record.media.empty? && !graphic_types.include?(record.media[0].file_type)
-      record.errors.add :media, "file is #{record.media[0].file_type}, not a graphic format we support."
+    if !record.media.empty? && !Concerto::ContentConverter.handles?(record.media[0].file_type)
```
bamnet: I don't think we should mix together both what we support natively and what we can convert here.  If the conversion library is scrubbed from the code everything should still work fine, right now we'd loose the validation if that happens.  I'd just OR the two different sources here.

app/models/graphic.rb
```diff
@@ -11,11 +10,16 @@ def validate(record)
...
   after_initialize :set_kind
+  before_save :convert_media
 
   #Validations
   validates :duration, :numericality => { :greater_than => 0 }
   validates :media, :length => { :minimum => 1, :too_short => "file is required." }
   validates_with GraphicValidator
+
+  def convert_media
+    self.media = Concerto::ContentConverter.convert(self.media)
```
bamnet: We should only call the convert if it can handle the content type.  It's confusing to always call convert when we're not actually expecting it to do anything.

lib/concerto/content_converter.rb
```diff
@@ -0,0 +1,103 @@
...
+      "application/vnd.oasis.opendocument.presentation",
+      "application/vnd.oasis.opendocument.spreadsheet",
+      "application/vnd.oasis.opendocument.text",
+      "application/vnd.openxmlformats-officedocument.presentationml.presentation",
+      "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+      "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
+      "image/x-eps"
+    ]
+
+    def self.handles? file_type
```
bamnet: Two style nits - most of our method defining code wraps the args in params.  If you've got a few seconds can you add a line or two of doc for methods too?  mac_addr.rb is an example of overkill, you don't need to be that verbose but somewhere in between would be great to help people follow along and make the generated docs useful (http://rubydoc.info/github/concerto/concerto/frames).

PR #769, opened 2013-12-13
Template import

### LINKED ISSUE #731: Allow template packages to be imported
Right now, the option to upload a template "package" is disabled and only the image and ill-defined "descriptor" fields are useful. We should change this to have a single upload field for a "package" - which will just be a JPEG and an XML descriptor file. This will be less confusing to users and allow the easy distribution of templates.

zr2d2:
adds template importing to close #731

### DIFF
--- a/app/controllers/templates_controller.rb
+++ b/app/controllers/templates_controller.rb
@@ -160,39 +160,13 @@ class TemplatesController < ApplicationController
   # template model.
   def import
     @template = Template.new
-
     archive = params[:package]
-    if archive.blank?
-      @template.errors.add(:base, t(:template_import_requires_archive))
-      respond_with(@template) do |format|
-        format.html { render :action => "new" }
-        format.xml  { render :xml => @template.errors, :status => :unprocessable_entity }
-      end
-    else
-      archive = archive.tempfile unless archive.is_a? Rack::Test::UploadedFile
-      xml_file = image_file = nil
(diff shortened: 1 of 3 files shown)


app/controllers/templates_controller.rb
```diff
@@ -156,34 +158,38 @@ def preview
...
       end
     else
-      begin
-        xml_data = xml_file.read
-        if !xml_data.blank? && @template.import_xml(xml_data)
-          @template.media.build({:key=>"original", :file => image_file})
-        end  
-      rescue REXML::ParseException
-        raise t(:template_import_error)
+      xml_file = image_file = nil
```
zr2d2: is there a better way to to do this?

app/controllers/templates_controller.rb
```diff
@@ -156,34 +158,39 @@ def preview
...
       end
     else
-      begin
-        xml_data = xml_file.read
-        if !xml_data.blank? && @template.import_xml(xml_data)
-          @template.media.build({:key=>"original", :file => image_file})
-        end  
-      rescue REXML::ParseException
-        raise t(:template_import_error)
+      archive = archive.tempfile unless archive.is_a? Rack::Test::UploadedFile
```
bamnet: Can you move the below import logic into the model?  I think we should have an import_archive(file) so we're not doing the heavy lifting in the controller here.

Ideally, the controller would be reduced to @template.import_archive(archive)

PR #784, opened 2013-12-20
fix #779 link to template

### LINKED ISSUE #779: screen#show header's template should be linked to the template
image
 Just like the owner is linked to the owner

mfrederickson:


app/views/screens/_screen_info.html.erb
```diff
@@ -13,7 +13,11 @@
     </div>
     <div class="span3">
       <p class="uppercase" style="margin-bottom: 0px;"><small><b><%= Screen.human_attribute_name(:template_id) %></b></small></p>
-      <p><%= screen.template.name unless screen.template.nil? %></p>
+      <% if !screen.template.nil? && can?(:read, screen.template) %>
+        <p><%= link_to screen.template.name, screen.template %></p>
+      <% else %>
+        <p><%= screen.template.name unless screen.template.nil? %></p>
```
zr2d2: There are two calls to `.nil?`, could we clean up the logic so there is only one check on the object Also, we should probably use `.blank?` instead

PR #837, opened 2014-01-17
Screen refresh capability

### LINKED ISSUE #808: Screens should periodically refresh templates
Right now, screens only pick up a template when they load. They should instead check for an updated (perhaps an emergency) template.

mfrederickson:
This should close #808 since it provides for the ability of a screen to refresh based on a header sent with content.


public/frontend_js/screen.js
```diff
@@ -71,6 +77,18 @@ goog.exportSymbol('concerto.frontend.Screen', concerto.frontend.Screen);
...
+/**
+ * Refresh the screen.
+ * Called typically when the content indicates a refresh is warranted
+ * (due to something like a template needing to be changed).
+ */
+concerto.frontend.Screen.prototype.refresh = function() {
+  var params = this.getQueryData();
+  var url = this.refresh_url + '?' + params.toString();
+  this.logger_.info('Refreshing screen from ' + url);
+  window.location.replace(url);
```
bamnet: Its not safe to assume thata screen is being used on a URL we control.  The frontend in Concerto is the primary consumer, but the frontend could also be embedded in a page by loading the js library and passing Screen a div creating a widget like you see @ http://concerto.rpi.edu (we call that guy the mini screen, and it was a total hack for v1).

mfrederickson: So what it sounds like you are saying is that we should not do a window.location.replace, but work from the screen's current div by wiping out it's content and starting from there...?

bamnet: I think that would be a good first approach, yea.

On Jan 18, 2014, at 1:15 AM, Marvin Frederickson notifications@github.com
wrote:

In public/frontend_js/screen.js:

> @@ -71,6 +77,18 @@ goog.exportSymbol('concerto.frontend.Screen', concerto.frontend.Screen);
>  concerto.frontend.Screen.prototype.logger_ = goog.debug.Logger.getLogger(
>      'concerto.frontend.Screen');
> 
> +/**
> - \* Refresh the screen.
> - \* Called typically when the content indicates a refresh is warranted
> - \* (due to something like a template needing to be changed).
> - */
>   +concerto.frontend.Screen.prototype.refresh = function() {
> -  var params = this.getQueryData();
> -  var url = this.refresh_url + '?' + params.toString();
> -  this.logger_.info('Refreshing screen from ' + url);
> -  window.location.replace(url);

So what it sounds like you are saying is that we should not do a
window.location.replace, but work from the screen's current div by wiping
out it's content and starting from there...?

—
Reply to this email directly or view it on
GitHubhttps://github.com/concerto/concerto/pull/837/files#r8986675
.

PR #863, opened 2014-01-26
pa should handle activity from plugins

mfrederickson:
The original recipient == owner check was preventing activity from being created in some instances, since it's intent was to prevent mailings, that logic was moved to the mailer's conditional.  Also enabled passing of custom key since a plugin's key gets pretty convoluted.

### DIFF
--- a/app/views/dashboard/_activities.html.erb
+++ b/app/views/dashboard/_activities.html.erb
@@ -5,7 +5,7 @@
     <% 
     begin
       a = render_activity activity 
-    rescue => e
+    rescue ActionView::MissingTemplate => e
       Rails.logger.error('public_activity view not found ' + e.message)
       a = nil
     end


app/views/dashboard/_activities.html.erb
```diff
@@ -1,8 +1,17 @@
 <% if !activities.empty? %>
   <ul class="list-stacked condensed">
   <% activities.each do |activity| %>
-    <% if File.exists?(Rails.root.join("app", "views", "public_activity", activity.key.split('.')[0], "_#{activity.key.split('.')[1]}.html.erb")) %>
-      <li><b><%= activity.created_at.strftime("%D, %l:%M") %>:</b> <%= render_activity activity %></li>
+    <%# if File.exists?(Rails.root.join("app", "views", "public_activity", activity.key.split('.')[0], "_#{activity.key.split('.')[1]}.html.erb")) %>
+    <% 
+    begin
+      a = render_activity activity 
+    rescue ActionView::MissingTemplate => e
```
mfrederickson, line 8: @augustf this isn't for catching deleted records, it's for when someone sets up some public activity in a plugin, and the plugin didn't define the activity's view correctly, or the plugin has been deactivated.  This should be a rare scenario.

mfrederickson, line 8: Actually when a plugin is deactivated its activity view still displays.

PR #873, opened 2014-02-02
Added Activity model, ConcertoConfig entry, and cron entry to achieve #871

### LINKED ISSUE #871: purge public activity older than x days
Add a setting for this.  0 days = keep forever.  Need class and method to delete records beyond this age, scheduled from lib/cron.rb to run once a day.

augustf:
@mfrederickson Does this seem like a reasonable way to resolve #871?

### DIFF
--- a/lib/cron.rb
+++ b/lib/cron.rb
@@ -21,12 +21,14 @@ module Clockwork
     Submission.delay.deny_old_expired
   end
   
-  every(1.day, 'Remove old public activity entries') do  
-    unless ConcertoConfig[:keep_activity_log].to_i == 0
-      activities =  PublicActivity::Activity.where("created_at > :days", {:days => ConcertoConfig[:keep_activity_log].to_i.days.ago}).destroy_all
-    end
-  end  
-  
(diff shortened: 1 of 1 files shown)


app/models/activity.rb
```diff
@@ -0,0 +1,3 @@
+class Activity < ActiveRecord::Base
```
bamnet: Does this model serve any purpose aside from being used to to destroy everything?  There must be a way to hook into the existing PublicActivity models to achieve this.

augustf: Nope;  If I had a better idea, I probably would have just merged this. I played a little bit of syntax bingo without success. As for looking into the public_activity code, grepping to see what class inherited from AR proved fruitless (you also don't want to know how they're extending AR). If we can't discover the proper syntax for latching into that table, this may be the easiest way of latching into our collection of activities directly.

augustf: Hrm-after another more detailed search, I think I found it. Seems one of my tests locally failed as PA hadn't been loaded (prior to the 2.1 symbol commit) via Bundler.

PR #874, opened 2014-02-02
Added Gemfile.local to make it easier to install unicorn

yakatz:
And other gems that are local


Gemfile
```diff
@@ -81,3 +81,8 @@ gem 'docsplit'   # for graphics and pdf, ppt conversion
 
 gem 'ice_cube'
 gem 'recurring_select', '~> 1.2.1rc3'
+
+# Load a local Gemfile if it exists
```
bamnet: Can you move this block up right above the group :assets stuff?  We should try and keep all the external Gemfile evals in the same area.

PR #925, opened 2014-05-20
Template import css

### LINKED ISSUE #869: be able to import a css file for an existing template

mfrederickson:
closes #869

### DIFF
--- a/app/controllers/templates_controller.rb
+++ b/app/controllers/templates_controller.rb
@@ -82,8 +82,8 @@ class TemplatesController < ApplicationController
     auth!
 
     # get a copy of the params and remove the files as we process them
-    tps = template_params
-    tps[:media_attributes].each do |k, media|
+    template_parameters = template_params
+    template_parameters[:media_attributes].each do |k, media|
       if !media.empty?
         # for the files that were uploaded, determine their media key based on their extension
         new_media = @template.media.build(media)
@@ -96,10 +96,10 @@ class TemplatesController < ApplicationController
         end
       end
       # remove the html file from the attributes so it is not processed in the in the update_attributes below
-      tps[:media_attributes].delete(k)
-    end if !tps[:media_attributes].nil?
+      template_parameters[:media_attributes].delete(k)
+    end unless template_parameters[:media_attributes].nil?
 
-    if @template.update_attributes(tps)
+    if @template.update_attributes(template_parameters)
       process_notification(@template, {}, process_notification_options({:params => {:template_name => @template.name}}))
       flash[:notice] = t(:template_updated)
     end


app/controllers/templates_controller.rb
```diff
@@ -41,6 +41,8 @@ def new
   def edit
     @template = Template.find(params[:id])
     auth!
+    # make sure that we only have two html file inputs, one for the graphic ("original") and one for the stylesheet ("css")
+    @template.media.reject! {|m| true}
```
zr2d2, line 45: I think `.reject! { |m| true}` will delete all associated media. I think we still want media to be associated on edit

app/controllers/templates_controller.rb
```diff
@@ -78,18 +80,31 @@ def create
...
     @template = Template.find(params[:id])
     auth!
-    # don't stomp on the key, set elsewhere
-    # @template.media.each do |media|
-    #   media.key = "original"
-    # end
 
-    if @template.update_attributes(template_params)
+    # get a copy of the params and remove the files as we process them
+    tps = template_params
```
zr2d2: could we give this a more readable name? Perhaps `template_parameters`?

app/controllers/templates_controller.rb
mfrederickson, line 45: Doesn't delete it, just removes it from the collection.

app/controllers/templates_controller.rb
```diff
@@ -78,18 +80,31 @@ def create
...
+    tps[:media_attributes].each do |k, media|
+      if !media.empty?
```
zr2d2, line 87: change to `unless media.empty?`

app/controllers/templates_controller.rb
mfrederickson: Could, I just didn't want to confuse it with "was it template_params or template_parameters?" so I just abbreviated it to "tps" so it was significantly different yet still somewhat indicative.

app/controllers/templates_controller.rb
mfrederickson, line 87: To me, this is more readily understandable, but if it rubs you the wrong way, go for it.

app/controllers/templates_controller.rb
```diff
@@ -78,18 +80,31 @@ def create
...
+        # for the files that were uploaded, determine their media key based on their extension
+        new_media = @template.media.build(media)
+        extension = (new_media.file_name.blank? ? nil : new_media.file_name.split('.')[-1].downcase)
+        new_media.key = (extension == "css" ? "css" : "original") unless extension.nil?
+
+        # mark any existing @template.media with this same key as replaced (obsolete)
+        @template.media.each do |m|
+          m.key = 'replaced_' + m.key if m.key == new_media.key and m != new_media #!m.new_record?
+        end
```
zr2d2, line 96: Probably could be `@template.media.each do { |m| m.key = "replaced_#{m.key}" unless m.new_record? }`

app/controllers/templates_controller.rb
```diff
@@ -78,18 +80,31 @@ def create
...
-    #   media.key = "original"
-    # end
 
-    if @template.update_attributes(template_params)
+    # get a copy of the params and remove the files as we process them
+    tps = template_params
+    tps[:media_attributes].each do |k, media|
+      if !media.empty?
+        # for the files that were uploaded, determine their media key based on their extension
+        new_media = @template.media.build(media)
```
zr2d2, line 89: could remove the parenthesizes. It's a style choice, but I do

app/controllers/templates_controller.rb
mfrederickson, line 96: Unless someone decides to upload two images or two CSS files.  This way the
last one wins.

app/controllers/templates_controller.rb
```diff
@@ -78,18 +80,31 @@ def create
...
+        extension = (new_media.file_name.blank? ? nil : new_media.file_name.split('.')[-1].downcase)
+        new_media.key = (extension == "css" ? "css" : "original") unless extension.nil?
+
+        # mark any existing @template.media with this same key as replaced (obsolete)
+        @template.media.each do |m|
+          m.key = 'replaced_' + m.key if m.key == new_media.key and m != new_media #!m.new_record?
+        end
+      end
+      # remove the html file from the attributes so it is not processed in the in the update_attributes below
+      tps[:media_attributes].delete(k)
```
zr2d2: could remove parenthesizes

app/controllers/templates_controller.rb
```diff
@@ -78,18 +80,31 @@ def create
...
+    end if !tps[:media_attributes].nil?
```
zr2d2: `end unless template_parameters[:media_attributes].nil?` is clearer

app/controllers/templates_controller.rb
zr2d2, line 45: So then the comment is erroneous?

mfrederickson, line 45: No. It is correct.  We need two records in the association since there can be two files uploaded.

zr2d2, line 45: But it doesn't make sure there are two files, as proclaimed. It summarily deletes the media objects

mfrederickson, line 45: The code below that makes sure there are two empty associations for the file inputs.

zr2d2, line 45: but `#edit` is just loading the template for display, I'm not seeing why it needs empty media objects

mfrederickson, line 45: I believe it is because the file inputs are bound to the media_attributes property of the template model.

Issue #938, opened 2014-05-29
Membership not destroyed when user is denied pending approval


ddbruce:
When a user requests to be added to a group but is denied, they do not show up in "Add users to group," as they should. Thanks to @gabe283 for finding this lucifer.

PR #940, opened 2014-05-29
Add interface to unblock previously denied memberships

gbprz:
#938


app/controllers/groups_controller.rb
```diff
@@ -32,6 +32,7 @@ def new
 
   def manage_members
     @group = Group.find(params[:id])
+    @denied = Membership.where(:group_id => params[:id], :level => 0)
```
bamnet: I would clean this up as `@denied = @group.memberships.denied`.

gbprz: @bamnet good call, just changed it

Issue #956, opened 2014-06-08
Google Analytics integration


bamnet:
V1 made it easy to integrate Google Analytics, V2 should do this as well.  @kench  volunteered to look into this I believe.

PR #957, opened 2014-06-10
[user-analytics] Added support for Google Analytics.

kench:
RELATED ISSUE : #956
RELEASE NOTES :
This commit adds support for the "Universal Analytics" flavor of GA (analytics.js).
Concerto respects the "Do-Not-Track" setting set by the user-agent.

### DIFF
--- a/config/initializers/19-analytics.rb
+++ b/config/initializers/19-analytics.rb
@@ -1,0 +1,13 @@
+Rails.logger.debug "Starting 19-analytics.rb at #{Time.now.to_s}"
+
+# Concerto user analytics configuration
+if ActiveRecord::Base.connection.table_exists? 'concerto_configs'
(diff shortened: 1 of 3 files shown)


config/initializers/19-analytics.rb
```diff
@@ -0,0 +1,13 @@
+Rails.logger.debug "Starting 19-analytics.rb at #{Time.now.to_s}"
+
+# Concerto user analytics configuration
```
augustf, line 3: @kench This would probably be better included in the initializer with the rest of the configs - no need for another one.

kench, line 3: I kept it a separate initializer to make it easier to remove out for any users that don't want Google Analytics support.

Also, I plan on refactoring it into a plugin (very soon) when I figure out better integration points in Concerto proper.

Issue #963, opened 2014-06-11
top two settings items missing labels


mfrederickson:
![image](https://cloud.githubusercontent.com/assets/473165/3247817/7a6d77dc-f188-11e3-8a6c-d50902edcda1.png)

PR #964, opened 2014-06-11
fix time field for #963

mfrederickson:
render time fields with proper ids and translations and change the order of the fields to a more natural progression (mix before max, start before end)...

### DIFF
--- a/config/initializers/02-concerto_config.rb
+++ b/config/initializers/02-concerto_config.rb
@@ -11,14 +11,14 @@ end
 if ActiveRecord::Base.connection.table_exists? 'concerto_configs'
   if ConcertoConfig.columns_hash.has_key?("plugin_id")
     # defaults
-    ConcertoConfig.make_concerto_config("default_upload_type", "graphic", :category => 'Content')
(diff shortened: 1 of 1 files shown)


app/views/concerto_config/show.html.erb
```diff
@@ -63,8 +63,9 @@
                   <%= label_tooltip "concerto_config", c.key.to_sym, t("activerecord.attributes.concerto_config.#{c.key.to_sym}"), :tip => c.description %>
                 </div>
               <% elsif c.value_type == "time" %>
-                <div class="timefield input-small">
-                  <%= text_field("start_time[time]", :id => "start_time_time", :maxlength => 20, :class => "timefield input-small", :value => ConcertoConfig[:content_default_start_time]) %>
+                <div class="input">
+                  <%= label_tooltip "concerto_config", c.key.to_sym, t("activerecord.attributes.concerto_config.#{c.key.to_sym}"), :tip => c.description %>
```
simplysoft, line 67: just a minor thing, but all other config do have the label_tooltip outside the `<div class="input">` seems to not have a negativ effect at the moment, but might produce problems in the future when the css changes

mfrederickson, line 67: Thanks, can't believe I missed that.

Issue #999, opened 2014-06-18
owner selection tweaks


mfrederickson:
The _Groups_ and _Users_ headers may need to be styled to stand out a little more than they currently do.
![image](https://cloud.githubusercontent.com/assets/473165/3321958/32d2969c-f740-11e3-8f74-4eba3c115d29.png)

When no groups match the filter, it looks like the headers collapse to inline, which is a little confusing.  It might be better to show "no matching groups" instead under the group header (same for the user, if no matching users).
![image](https://cloud.githubusercontent.com/assets/473165/3321975/6bd6a2b2-f740-11e3-85b7-e9fd72c1ff04.png)

PR #1007, opened 2014-06-19
Owner selection

gbprz:
#999

### DIFF
--- a/app/views/elements/_owner_select.html.erb
+++ b/app/views/elements/_owner_select.html.erb
@@ -21,32 +21,32 @@
         <% if !options["Groups"].nil? %>
           <% if options["Groups"].length > 0 %>
             <h4 class="default-padding" style="text-align: center"> Groups </h4>
-          <% end %>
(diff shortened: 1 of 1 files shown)


app/views/elements/_owner_select.html.erb
```diff
@@ -19,9 +19,10 @@
       <ul class="filterable selector-list">
 
         <% if !options["Groups"].nil? %>
-          <b class="default-padding" style="text-align: center"> Groups </b>
+          <% if options["Groups"].length > 0 %>
+            <h4 class="default-padding" style="text-align: center"> Groups </h4>
             <% options["Groups"].each do |group| %>
-              <li class = "link-hl default-padding" style="padding: 10px"
+              <li id="ownerItem" class = "link-hl default-padding" style="padding: 10px"
```
simplysoft, line 25: `ownerItem` should be rather a class instead of an id. Id's are only allowed to appear once in the whole document. Further below in Line 41 there is the same problem

PR #1019, opened 2014-06-20
move feed selection to separate partial so form hook can display above i...

mfrederickson:
...t

### DIFF
--- a/app/views/contents/_edit_form.html
+++ b/app/views/contents/_edit_form.html
@@ -9,6 +9,7 @@
     </div>
     <div class="span8">
       <%= render "form_middle", {:form => f} %>
+      <%= render "form_submit_feeds", {:form => f}) %>
     </div>
   </div>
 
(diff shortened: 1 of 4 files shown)


app/views/contents/_edit_form.html
```diff
@@ -9,6 +9,7 @@
     </div>
     <div class="span8">
       <%= render "form_middle", {:form => f} %>
+      <%= render "form_submit_feeds", {:form => f}) %>
```
bamnet, line 12: This prohibits a content type from doing something special with feed selection.  Is there a bug this fixes?

mfrederickson, line 12: No, when I split the feed selection out of the form_middle to allow plugin's to add additional fields above the feed selection, I merely added the feed selection into the edit form also.  

How does this prevent a content type from doing something special with feeds?  Can't it just override the form_submit_feeds just like it does the form_middle?

mfrederickson, line 12: Although, I don't think the edit form is referring to any of the overridden partials...

PR #1020, opened 2014-06-20
extend plugin and plugin info to allow for extending models

### LINKED ISSUE #1014: Add possibility to extend concerto models through PluginInfo
It might be useful that concerto provides a way to extend concerto models through the PluginInfo. This is e.g. needed when a plugin adds his own models and would like to extend the existing ones with `has_many` or other relations.

Right now something like that can be achieved by using rails functionality, but this has the downside that this is always run, even if the plugin is not enabled:

```
module YourConcertoPlugin
  class Engine < ::Rails::Engine
    config.to_prepare do
      Screen.class_eval { include YourConcertoPlugin::Concerns::SomeExtension }}
    end 
  end
end
```

I think it would be great if the PluginInfo would support a way registering [concerns](http://api.rubyonrails.org/classes/ActiveSupport/Concern.html) for the concerto models, similar to how add_controller_hook works. Something like
`extend_model Screen, YourConcertoPlugin::Concerns::SomeExtension`

simplysoft:
fixes #1014

Example: 

```
module ConcertoCustomPlugin
  class Engine < ::Rails::Engine
    isolate_namespace ConcertoStatistics
    engine_name "statistics"

    def plugin_info(plugin_info_class)
      @plugin_info ||= plugin_info_class.new do
          extend_model(Screen, ConcertoCustomPlugin::Concerns::AddCustomizationToScreen)
        end
      end
    end
  end
end
```

```
module ConcertoCustomPlugin
  module Concerns
    module AddCustomizationToScreen
      extend ActiveSupport::Concern
      included do
        has_many :things, :class_name => 'ConcertoCustomPlugin::Thing', :dependent => :destroy
      end
    end
  end
end
```

### DIFF
--- a/config/initializers/12-plugins.rb
+++ b/config/initializers/12-plugins.rb
@@ -23,6 +23,10 @@ if ActiveRecord::Base.connection.table_exists? 'concerto_plugins'
       end
     end
   end
+
+  Rails.application.config.to_prepare do
+    ConcertoPlugin.extend_models
+  end
 end
 
 Rails.logger.debug "Completed 12-plugins.rb at #{Time.now.to_s}"
--- a/app/models/concerto_plugin.rb
+++ b/app/models/concerto_plugin.rb
@@ -148,6 +148,22 @@ class ConcertoPlugin < ActiveRecord::Base
     end
   end
 
+  # Extends models by including modules (preferably ActiveSupport::Concern)
+  def self.extend_models
+    ConcertoPlugin.enabled.each do |plugin|
+      info = plugin.plugin_info
+      next if info.nil? || info.model_extensions.nil?
+
+      info.model_extensions.each do |model, extensions|
+        model.class_eval do
+          extensions.each do |extension|
+            include extension
+          end
+        end
+      end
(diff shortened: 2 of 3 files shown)


config/initializers/12-plugins.rb
```diff
@@ -23,6 +23,10 @@
       end
     end
   end
+
+  Rails.application.config.to_prepare do
+    ConcertoPlugin.extend_models
+  end
```
mfrederickson, line 29: Should this happen before the initialize_plugins above?

simplysoft, line 29: hm.. good question. 
The current order would not allow you to use your extended models inside the plugins init block. Not sure if that would be a valid use case. 

But even moving the the `Rails.application.config.to_prepare` up wouldn't really solve this problem, because the  block is executed is later (As far as I know after all initializers are run and before the first request)

Even if we flip the order, this would mean

mfrederickson, line 29: Oh yeah, I didn't think about the later execution.

PR #1036, opened 2014-06-24
add configurable paging size

### LINKED ISSUE #1032: inconsistent paging size
When browsing content I see more than eight items under active content with no paging indicators-- but when I click on a user name to see their submitted content, the list is paginated with eight items per page.  I'd opt for a page size of 50 or so (or configurable)...

mfrederickson:
closes #1032

### DIFF
--- a/config/initializers/02-concerto_config.rb
+++ b/config/initializers/02-concerto_config.rb
@@ -52,6 +52,7 @@ if ActiveRecord::Base.connection.table_exists? 'concerto_configs'
     ConcertoConfig.make_concerto_config("keep_activity_log", "90", :value_type => "integer", :value_default => "90", :category => 'System', :description => 'Days to keep activity log for (where 0 is forever)')
     ConcertoConfig.make_concerto_config("footer_html", "", :value_type => "text", :category => 'System', :description => 'HTML to display on the footer of every page.')
     ConcertoConfig.make_concerto_config("secret_token", "", :value_type => "string", :value_default => "", :hidden => "true", :category => 'System')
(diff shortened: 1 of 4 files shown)


config/initializers/02-concerto_config.rb
```diff
@@ -52,6 +52,7 @@
     ConcertoConfig.make_concerto_config("keep_activity_log", "90", :value_type => "integer", :value_default => "90", :category => 'System', :description => 'Days to keep activity log for (where 0 is forever)')
     ConcertoConfig.make_concerto_config("footer_html", "", :value_type => "text", :category => 'System', :description => 'HTML to display on the footer of every page.')
     ConcertoConfig.make_concerto_config("secret_token", "", :value_type => "string", :value_default => "", :hidden => "true", :category => 'System')
+    ConcertoConfig.make_concerto_config("items_per_page", "50", :value_type => "integer", :value_default => "50", :category => 'System')
```
bamnet, line 55: This feels like config creep to me.  Are there users who need to change this?

augustf, line 55: Well I could see users in larger instances wanting to change items per page, but it should be a client-side setting, not a server config.

mfrederickson, line 55: Do we have a mechanism for that? (Other than a column in the user table?)

augustf, line 55: Actually I was more thinking on a per-session basis. Think of most websites you visit where there's a widget to set items per page.

mfrederickson, line 55: Ah... and the nicer ones persist it for you.  It could get old having to change it each time you visit.

augustf, line 55: Well if we store it in the session object, that could persist for a good while...

PR #1043, opened 2014-06-24
Ldap

mfrederickson:
Provides LDAP authentication.  To use-- simply set the LDAP Server in settings and restart your webserver.

### DIFF
--- a/config/initializers/02-ldap_authenticatable.rb
+++ b/config/initializers/02-ldap_authenticatable.rb
@@ -17,14 +17,19 @@ if ActiveRecord::Base.connection.table_exists? 'concerto_configs' and !ConcertoC
         end
 
         def self.verified?(user, password)
-          ldap = Net::LDAP.new
-          if ActiveRecord::Base.connection.table_exists? 'concerto_configs'
-            ldap.host = ConcertoConfig[:ldap_host] 
-            ldap.port = ConcertoConfig[:ldap_port] 
+          begin
+            ldap = Net::LDAP.new
+            if ActiveRecord::Base.connection.table_exists? 'concerto_configs'
+              ldap.host = ConcertoConfig[:ldap_host] 
+              ldap.port = ConcertoConfig[:ldap_port] 
+            end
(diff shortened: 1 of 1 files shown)


config/initializers/02-ldap_authenticatable.rb
```diff
@@ -0,0 +1,51 @@
...
+        def email
+          params[:user][:email]
+        end
+
+        def password
+          params[:user][:password]
+        end
+
+        def user_data
+          {:email => email, :password => password, :password_confirmation => password}
```
kench, line 44: We shouldn't be storing the user's password from LDAP in the Concerto database.

config/locales/devise.en.yml
```diff
@@ -16,6 +16,7 @@ en:
       timeout: "Your session expired, please sign in again to continue."
       unauthenticated: "You need to sign in or sign up before continuing."
       unconfirmed: "You have to confirm your account before continuing."
+      ldap: "Unable to verify account via LDAP"
```
kench, line 19: "Unable to connect to LDAP server."

config/initializers/02-ldap_authenticatable.rb
mfrederickson, line 44: We don't

On Aug 2, 2014, at 11:48 AM, Kenley Cheung notifications@github.com wrote:

In config/initializers/02-ldap_authenticatable.rb:

> -            Rails.logger.error("Unable to authenticate using LDAP - #{e.message}")
> -            verified = false
> -          end
> -          return verified
> -        end
>   +
> -        def email
> -          params[:user][:email]
> -        end
>   +
> -        def password
> -          params[:user][:password]
> -        end
>   +
> -        def user_data
> -          {:email => email, :password => password, :password_confirmation => password}

We shouldn't be storing the user's password from LDAP in the Concerto
database.

—
Reply to this email directly or view it on GitHub
https://github.com/concerto/concerto/pull/1043/files#r15730951.

PR #1100, opened 2014-08-13
Add fields for content search and refactor content filtering

gbprz:


app/models/content.rb
```diff
@@ -190,35 +190,28 @@ def after_add_callback(unused_submission)
...
-      query_conditions[:type] = params[:type] if params[:type]
-      if query_conditions.empty?
-        # No filters are specified
-        filtered_contents = Content.all
-      else
-        # User and/or type filters are specified
-        filtered_contents = Content.all(:conditions => query_conditions)
-      end
+    # Retrieve all contents
+    filtered_contents = Content.all
```
simplysoft: It would be better to harness `ActiveRecord::Relation` and use `.where` below so that only relevant content is loaded from the database (let the db do the filtering)

gbprz: Good point, it will scale much better than this solution too, especially if the search parameters are  very specific and we don't have to sift through every piece of content in the database.

PR #1124, opened 2014-09-23
Visibility & i18n fixes

simplysoft:


app/views/devise/registrations/new_first_admin.html.erb
```diff
@@ -3,7 +3,7 @@
     <div class="default-padding" style="padding: 40px 12px; text-align: center;">
       <%= image_tag("elements/logomark.png", :height => 175, :alt => "") %>
       <br /><br />
-      <h1 style="font-size: 5em; letter-spacing: -0.025em;">Welcome to Concerto.</h1>
+      <h1 style="font-size: 5em; letter-spacing: -0.025em;">Welcome to <%= t('concerto') %>.</h1>
```
zr2d2, line 6: Concerto is a brand name, and I don't think we should translate it. Also, translation strings would need a corresponding definitions in the apropriate `config/locals/LANGUAGE_CODE.yaml` file

simplysoft, line 6: It already exists: https://github.com/concerto/concerto/blob/master/config/locales/en.yml#L22
so it might be worth using it (or we should completely remove it everywhere)

PR #1128, opened 2014-09-27
Rails 4

augustf:


config/concerto.yml
```diff
@@ -2,4 +2,4 @@ automatic_bundle_installation: true
 automatic_database_installation: true
 compile_production_assets: true
 airbrake_enabled_initially: true
-bundle_install_options: "--path vendor/bundle"
```
simplysoft, line 5: Was this changed on purpose? If so, what was the motivation?

config/initializers/02-concerto_config.rb
```diff
@@ -8,6 +8,9 @@
   Rails.logger.debug "Socket error in trying to determine hostname: #{e}"
 end
 
+require 'yaml'
+concerto_base_config = YAML.load_file("./config/concerto.yml")
+
```
simplysoft, line 13: `concerto_base_config` is not used (yet?). Shall we keep or remove it?

config/concerto.yml
zr2d2, line 5: I changed it during development for my convenience, but it shouldn't have been merged like that

app/models/concerto_plugin.rb
```diff
@@ -262,7 +260,7 @@ def find_engine
       gem_basename = Pathname(Gem.loaded_specs[gem_name].full_gem_path).basename
       # Then match the name we've got to the name of an engine -
       #    which should have its Module Name (aka paydirt)
-      Rails::Application::Railties.engines.each do |engine|
+      ::Rails::Engine.subclasses.map(&:instance).each do |engine|
```
simplysoft, line 263: This change is likely unnecessary as `Rails::Application::Railties.engines` does exactly the same (see https://github.com/rails/rails/blob/4-0-stable/railties/lib/rails/engine/railties.rb#L12-14)

If it was not replaced for good reason (maybe caching?),  `Rails::Application::Railties.engines` should be preferred because of better future robustness

simplysoft, line 263: @khanh can you clarify the motivation of this change?

zr2d2, line 263: you point to 4.0-stable, but i suspect this may not work in rails 4.1

PR #1141, opened 2014-10-31
Translation pluralization issue on screen show page

gbprz:


config/locales/en.yml
```diff
@@ -47,7 +47,7 @@ en:
   other_model: Other %{model}
   parameter:
     one: parameter
-    many: parameters
+    other: parameters
```
zr2d2, line 50: i keep using the wrong word. Thank you

Issue #498, opened 2013-06-03
Add timezone field to screens


augustf:
Branching out of the discussion on #481, we should consider adding a timezone field for screens, such that screens in remote locations can override the system time zone set in ConcertoConfig.

augustf:
I've had to revert 16fc86264dc712a205e20ac82df2d6b192d9200d, as I'm unable to run further migration

PR #1152, opened 2014-12-16
Send ISO 8601 date for content#create actions.

### LINKED ISSUE #1143: Hardcoded time formats in content model
The content model does parse the time using some hardcoded time formats.
- https://github.com/concerto/concerto/blob/master/app/models/content.rb#L99
- https://github.com/concerto/concerto/blob/master/app/models/content.rb#L111

The time format should either be fetched from the translations or some other way of transferring date and time from the UI to the controllers should be used (e.g. timestamp).

Are there any insights why the current way of doing it was necessary?

kench:
MOTIVATION & SCOPE : Closes #1143
TESTING : Check Travis CI and pull down code to check.
REVIEWERS : @zr2d2, @simplysoft, @bamnet

### DIFF
--- a/app/models/content.rb
+++ b/app/models/content.rb
@@ -31,7 +31,7 @@ class Content < ActiveRecord::Base
   # By default, only find known content types.
   # This allows everything to keep working if a content type goes missing
   # or (more likely) gets removed.
-  default_scope { where(:type => Content.all_subclasses.collect { |s| s.name })}
+  default_scope { where(:type => Content.all_subclasses.collect { |s| s.name }) }
 
   # Easily query for active, expired, or future content
   # The scopes are defined as class methods to delay their resolution, defining them as proper scopes
@@ -93,7 +93,7 @@ class Content < ActiveRecord::Base
   # times between 1.9.x and 1.8.x.
   def start_time=(_start_time)
     if _start_time.kind_of?(String)
-      write_attribute(:end_time, Time.parse(_start_time)
+      write_attribute(:end_time, Time.parse(_start_time))
     else
       write_attribute(:start_time, _start_time)
     end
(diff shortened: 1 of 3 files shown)


app/models/content.rb
```diff
@@ -92,12 +92,8 @@ def is_denied?
   # Otherwise, just set it like normal.  This is a bit confusing due to the differences in how Ruby handles
   # times between 1.9.x and 1.8.x.
   def start_time=(_start_time)
-    if _start_time.kind_of?(Hash)
-      return if _start_time[:date].empty?
-      #write_attribute(:start_time, Time.parse("#{_start_time[:date]} #{_start_time[:time]}").to_s(:db))
-      # convert to time, strip off the timezone offset so it reflects local time
-      t = DateTime.strptime("#{_start_time[:date]} #{_start_time[:time]}", "%m/%d/%Y %l:%M %p")
-      write_attribute(:start_time, Time.zone.parse(Time.iso8601(t.to_s).to_s(:db)).to_s(:db))
+    if _start_time.kind_of?(String)
```
bamnet: I think you actually want to knock out the else clause below this if instead of the if Hash logic.  If hash, elseif string.  _start_time will either be a hash (if someone has built their own completely custom content form) or a string (if they are using your updated JS logic like they should)... there's no more else scenario for a non-string non-hash.

augustf:
@kench Anything new with this work on dates?

augustf:
@simplysoft Is there anything anyone on the team can do to assist with #1143 and this work?

simplysoft:
Clarifications on how timezone are expected to be handled would help a lot. In particular the question is if  the content duration time is meant to be in client time / browser zone or server time zone. 
And we would have to verify if everywhere were those times are used the applicable timezone is used

kench:
[ISO 8601](http://en.wikipedia.org/wiki/ISO_8601) can encode the UTC offset, but time should be stored in [UTC](http://en.wikipedia.org/wiki/Coordinated_Universal_Time) to avoid ambiguity (not server or client timezone).

augustf:
I think @simplysoft is asking more about end user expectations (regardless of internal representation). If a user is in one time zone and a screen in another, what are user expectations? I would tend to think that when a user enters the time they want content to run, they're thinking in their local terms (client time). The server time doesn't seem relevant, and we don't yet know what screens the content may be going to. #498 deals with screen timezone issues.

PR #1162, opened 2014-12-23
Housekeeping

simplysoft:
As the year-end is approaching, let's join efforts and do some housekeeping :)

This PR is a collection of various fixes / cleanups (see individual commits) that did not yet found their way back upstream

### DIFF
--- a/app/controllers/frontend/contents_controller.rb
+++ b/app/controllers/frontend/contents_controller.rb
@@ -22,7 +22,7 @@ class Frontend::ContentsController < ApplicationController
 
     shuffle_config = FieldConfig.get(@screen, @field, 'shuffler') || DEFAULT_SHUFFLE
     shuffler_klass = FrontendContentOrder.load_shuffler(shuffle_config)
-    session_key = "frontend_#{@screen.id}_#{@screen.template.id}_#{@field.id}".to_sym
+    session_key = "frontend_#{@screen.id}_#{@screen.template.id}_#{@field.id}_#{shuffler_klass}".to_sym
     shuffler = nil
     count = 1
 


app/assets/javascripts/rangeinput.js
```diff
@@ -61,9 +61,6 @@
 
     $.fn.drag = function(conf) {
 
-        // disable IE specialities
-        document.ondragstart = function () { return false; };
-
```
zr2d2, line 66: what versions of IE was this for and why remove it?

app/controllers/frontend/contents_controller.rb
```diff
@@ -22,7 +22,7 @@ def index
 
     shuffle_config = FieldConfig.get(@screen, @field, 'shuffler') || DEFAULT_SHUFFLE
     shuffler_klass = FrontendContentOrder.load_shuffler(shuffle_config)
-    session_key = "frontend_#{@screen.id}_#{@screen.template.id}_#{@field.id}".to_sym
+    session_key = "frontend_#{@screen.id}_#{@screen.template.id}_#{@field.id}_#{shuffler_klass}".to_sym
```
zr2d2, line 25: why does this session key depend on the type of shuffler used?

app/assets/javascripts/rangeinput.js
simplysoft, line 66: Not sure, I assume IE 8 or earlier, we could not track the origin of this code. We have tried it with latest IE, and this code does not seem to have any effect (anymore).

This code is actually quite nasty, it overwrites the ondrag callback on the document, and because it is returning false, it means that dragging will never start (for code other than the range input). 

You can verify this by add a simple draggable code e.g. from http://jqueryui.com/draggable/ . As soon as you have a rangeinput on the same page (i.e. duration slider), the dragging won't work anymore, as dragging will never start

app/controllers/frontend/contents_controller.rb
simplysoft, line 25: please read the corresponding commit message: "make suffler class part of session key to prevent crash when switching shufflers that have different cache data format"

we have tested some different shuffler implementation that do not return an array (but a hash) when saving to the session (`save_session`). When we then switched the shufflers, the session data did restore the wrong store (array instead of hash or vice versa)

Issue #1135, opened 2014-10-19
i18n translations are duplicated


simplysoft:
It looks like that a lot of the translations are duplicated. All translations seem to exist in the the config/locales/en.yml file, but the translations are duplicated in various smaller files for individual components (e.g. config/locales/views/activity_mailer/en.yml)

Either we should remove the global en.yml fie or get rid of the various smaller individual files.

simplysoft:
A year passed, and I have new insights:

It is better to have component specific yml files. It has the risk of duplication, but the duplication is intentional. Every specific yml file defines a translation context and even if the two texts in two different contexts are the same, this does not mean that they are equal. 

I think we developer tend to remove duplication as much as possible, but in the translation case this is actually harmful. Context matters and in almost all cases. 

And it turns out that there are solution for this "duplication problem" in the translation space. Professional translation works with translation memory / fuzzy searching, so the translator can decide on case by case if he should use the same term or another. So the overhead of having multiple translation keys that have the same text is significantly reduced

I have only one open question left:
How should we deal with the non-english translation? Should we also make the effort and split them into the individual files, or is it ok to have them in one big file. As far as I know management of one big file is much simpler with a lot of translation tools

simplysoft:
We have just encountered one example of harmful de-duplication developer tend to perform:
https://github.com/concerto/concerto/blob/master/app/views/public_activity/membership/_create.html.erb

This codes assume one certain grammatical structure of the language, if this does not match, it is almost not possible to translate it.

The proper solution to fix this is to create translation strings for every combination of the possible text snippets. It might feel counterintuitive at first, but one gets used to it :)

Issue #929, opened 2014-05-22
Translation Tool


simplysoft:
We are looking into translating concerto into german. It would be great if we could use a translation tools like https://www.localeapp.com (free for public projects)
This would ease development and the work of the translators.

So if a responsible person would be kind enough to create an account, that would be great.

simplysoft:
Could someone of the team (@augustus ?) apply for a free account at https://crowdin.com/page/open-source-project-setup-request

We would like to use it to translate concerto to german. As far as we have tested, this service provider provides the best solution (includes translation memory, automatic machine translations, etc)

PR #1163, opened 2015-01-03
Cleanup translation

simplysoft:
It is a mess, but I tried my best.

See the individual commits to keep track of the changes, I tried to make it as auditable as possible. 
The deduplication was performed by some hand crafted scripts / extension to i18n-tasks, so there should be no human errors in there. The script removed all keys in the main en.yml if there was another translation somewhere in the subdirectories

This PR is an effort for solving #1135 and supporting #929


zr2d2:
Thanks. I'll look at it tonight. I made a commit a while ago that added duplication, and never Tom the time to remove it. Didn't seem that urgent

simplysoft:
Great. 
Luckily there were not much inconsistencies, you can find them here: https://gist.github.com/simplysoft/1162b3c40bacfbc4e562

zr2d2:
nice work! I've been waylayed at work as of late, but this is a good addition

PR #1164, opened 2015-01-04
Increase i18n coverage

simplysoft:
:fork_and_knife: This PR built on top of #1163 and has the goal to make every string translatable. Full coverage is not yet achieved, but we will update this PR continuously.

:warning: Due to the process, the existing non-default translations (fr, nl, he) will regress slightly because translation keys can be restructured and it is too much work to perform all changes on all language. 
We propose that if necessary the missing translations will be fixed at the very end of the work. Before we started there were already quite some translations missing for the different languages, and the work will introduce new strings that need to be translated (which at the moment are hardcoded in model/controller/views). 

For the english language we keep track of all changes, so no translation should be lost, and new introduced translation keys will be translated to english.

Feel free to join the effort and help reviewing the changes, as this is quite some repetitive work that is prone to careless mistakes 

**TODO:**
- [x] refactor duplicate :prohibited_save error message generation (kinds/_form.html.erb, templates/_create_form.erb, ...)
- [x] refactor public activity file so that they don't try to generate human texts (which badly fails for foreign languages with different sentence structure)
- [x] provide method to translate javascript
- [x] translate javascript strings


simplysoft:
It is finished :sweat:
It would be great if you can regression test this PR. We did already test the english translation for regression, but might very well be that we overlooked some things.

Once this PR is merged, we will be able to contribute a complete german translation: https://www.localeapp.com/projects/7725

We should also device what to do with the other languages (fr, nl, he)

augustf:
@simplysoft Do you have a list of the translations that would be missing from the non-English locales? it would be good just to have a sense of the size of regression we're creating as well as a starting point for those looking to do supplemental translation.

simplysoft:
@augustf not yet, but we will try to create one. The problem with the current tool (localeapp.com) is that there is no base language configuration possible, so outdated keys from fr, nl or he will be considered new strings to be translated for english and german, so it is not that easy to create that list. (and it would screw our existing localeapp project we used for german translation (https://www.localeapp.com/projects/7725), so we were reluctant to do it)

Would be great if you can request an account at https://crowdin.com/page/open-source-project-setup-request so that we can centralize efforts to translate concerto and have one common language repository. Then we would be able to upload all the existing translation and see how much is outstanding.

My gut feeling is that there were about 50 keys changed and 100 keys were newly added.

augustf:
Rather than take on another new toolchain, I may investigate a ruby script (mine or someone else's) to get a list of the missing translations outside of English and German (from a programming perspective, comparing the keys of 2 YML files should be trivial), contact the translators, and ask them if they'd like to do an update. It would be unfair to folks outside of our two languages to have their Concerto broken without at least a pathway to fixing it.

simplysoft:
We will try to create such a list (i18n-tasks should be able to do this)

Crowdin would be for translating the missing things. Yes you can also do it via a text editor, but it is tedious and error prone. And crowdin would help in keeping the languages in sync. That's a problem that this project experienced already. The translations were never updated after their first commit (even though new strings were introduced).

Wouldn't it be nice that the translator can just go to a website and enter the missing translations and we could then just automatically pull the translations from the system? This would make the work easier for both the translators and for us the developers.  

And I think the issue about the "concerto being broken" for the other language is not the much of the problem. Until now it was never possible to properly use concerto translated into a foreign language (estimated 15% of the system was not possible to be translated, there were a lot of hardcoded english strings), so if its a bit broken or a bit more broken does not matter much (at least in my opinion)

simplysoft:
Please find here a list of the missing translation: https://gist.github.com/simplysoft/c4381d2bf9c879563088

You can ignore he, it was never really translated

augustf:
Given this list, I think it's reasonable to merge these changes. As we contact our French and Dutch contributors, it might be helpful to, in the files, indicate where there work is needed somehow.

augustf:
Hrm-I may have erred in merging #1163 before looking at this. It appears this branch is non-fast-forward-mergeable now. I'll look into it...

simplysoft:
#1163 is just a cleanup, so won't introduce any regression. Now it should be mergable

simplysoft:
Great, thanks for merging :+1: 

Before releasing it would be great if everybody could do some regression testing. We have done quite some testing so far and e.g. have found nasty things like 0481c04dfb04f8435d2d0ae790cc730acd9a36d2 where translated strings are used in code other then just displaying.

app/views/concerto_plugins/_plugins_navigation.html.erb
```diff
@@ -7,7 +7,7 @@
     </li>
     <li class="active">
       <%= link_to concerto_plugins_path do %>
-          <i class="icon-bolt"></i> <%= t :manage_model, :model => ConcertoPlugin.model_name.human.pluralize %>
+          <i class="icon-bolt"></i> <%= t :manage_model, :model => ConcertoPlugin.model_name.human(count: 42) %>
```
zr2d2, line 10: this is a little better, but still always the same (probably always other: in the translation count), could we use ConcertoPlugin.count?

app/views/concerto_plugins/edit.html.erb
```diff
@@ -1,7 +1,7 @@
 <section class="viewblock">
   <header class="viewblock-header">
     <h1 class="default-padding">
-      <%= link_to t(:manage_model, :model => ConcertoPlugin.model_name.human.pluralize), concerto_plugins_path %> > <%= @concerto_plugin.name.titleize %>
+      <%= link_to t(:manage_model, :model => ConcertoPlugin.model_name.human(count: 42)), concerto_plugins_path %> > <%= @concerto_plugin.name.titleize %>
```
zr2d2, line 4: same here

app/views/concerto_plugins/index.html.erb
```diff
@@ -8,7 +8,7 @@
       </div>
     </div>
     <h1 class="default-padding">
-      <%= t :manage_model, :model => ConcertoPlugin.model_name.human.pluralize %>
+      <%= t :manage_model, :model => ConcertoPlugin.model_name.human(count: 42) %>
```
zr2d2, line 11: and here

PR #1171, opened 2015-01-24
German translation

simplysoft:
and some cleanups for easier language handling (non-english languages have each one file)

The translation status can be viewed at https://crowdin.com/project/concerto


augustf:
I'm not sure combining all our translations into a single unmanageable file is a very good idea. I imagine it makes web tools like CrowdIn work, but it takes away the modularity of our translation structure and removes any sense of which translations go where in the application's structure. @zr2d2 Any opinions on this?

simplysoft:
I think its easier to manage if for translations its one file. The key is that the big file we don't have to manage, whereas the 25 individual files we would have to.
Because translations always are behind the original language it it a hassle to update always 100 (25 files * 4 languages) individual files. This would only work with automation.
For the english part though we should keep the structure as we have it now. 

Actually crowdin is one of the few tools that supports having the files split up into multiple files, and in the initial try I did upload it them individually. But this also means that a translator has to select all the 25 files and then translate them, and we have to go through 25 files and approve translations. 
And the dutch language was already in one bug chunk, so we would first have to split this file again.

If you think it is worth splitting also the translations, then somebody should make an effort and write the rules that define in which files translations belong (we gave it a try for 5 minutes and then gave up). We have already tool support for this with i18n-tasks:
https://github.com/glebm/i18n-tasks#pattern-router

augustf:
I see what you mean. In my mind, the main virtue of separate files is that a file like views/screens/de.yml is the German strings that appear in the screens views in app/views/screens/.

augustf:
After conferring with @zr2d2 , we've decided to accept this as the benefits of separate files initially were minor.

zr2d2:
to clarify, I see separate file beneficial only in the way @augustf described, where a developer is manually mapping the keys back to the application structure. Since the non-English locals are being used programmaticaly and not by hand, I see no need to keep them seperate

PR #1177, opened 2015-02-02
Update _form.html.erb

romoloman:
The previous patch renders a blank page.

### DIFF
--- a/app/views/feeds/_form.html.erb
+++ b/app/views/feeds/_form.html.erb
@@ -35,7 +35,7 @@
         <div class="clearfix">
           <%= f.label :group %>
           <div class="input">
-            <%= f.collection_select :group_id, Group.accessible_by(current_ability), :id, :name %>
+            <%= f.collection_select :group_id, Group.all, :id, :name %>
           </div>
         </div>
         


augustf:
This definitely seems to fix this issue on the show page -- but I'm unable to understand either why this fixes the issue or why the other page changed in 8f97ceae16c73dbb127db922bdd85de5bc88f3cd is working. If we can't get to the bottom of this, I'll simply revert 8f97ceae16c73dbb127db922bdd85de5bc88f3cd , as it seems to have caused the regression.

romoloman:
I suppose that current_ability doesn't allow to read group list that's why accessible_by fails
The other part of 8f97cea seem to work properly.

augustf:
Talking with @mikldt , the problem may be related to the Groups not all being readable and a better approach than listing all the groups (which the user shouldn't have access to) is to use the auth! function in the controller for the groups array.

simplysoft:
auth! should do the same thing as accessible_by, so the real problem might be in the ability permission rules. 
We are using some modified rules, it can be that the original rules are not complete.

simplysoft:
The root cause are the ability rules for when public concerto is enabled. we are working on a fix

Issue #1174, opened 2015-01-27
500 error on /feeds/new page


snoop168:
Getting a 500 error. Other pages seem to load fine. my apache error.log file shows:

```
App 3622 stderr: /webapps/concerto/lib/command_check.rb:6: warning: Insecure world writable dir /webapps/concerto/vendor/bundle/ruby/1.9.1/bin in PATH, mode 040777
App 3622 stderr: /usr/local/bin/bundle:23:in `load'
App 3622 stderr: : cannot load such file -- /usr/lib/ruby/bin/bundle (LoadError)
App 3622 stderr: 	from /usr/local/bin/bundle:23:in `<main>'
```
Is it a problem with my bundler? the path /usr/lib/ruby/bin/ does not exist on my machine.

![image](https://cloud.githubusercontent.com/assets/2731594/5918139/4bbb64ba-a5f3-11e4-8c0b-1eb0ea5b605e.png)

zr2d2:
I would start by changing /webapps/concerto/vendor/bundle/ruby/1.9.1/bin
with `chmod go-w /webapps/concerto/vendor/bundle/ruby/1.9.1/bin`

Getting a 500 error. Other pages seem to load fine. my apache error.log
file shows:

App 3622 stderr: /webapps/concerto/lib/command_check.rb:6: warning:
Insecure world writable dir
/webapps/concerto/vendor/bundle/ruby/1.9.1/bin in PATH, mode 040777
App 3622 stderr: /usr/local/bin/bundle:23:in `load'
App 3622 stderr: : cannot load such file -- /usr/lib/ruby/bin/bundle (LoadError)
App 3622 stderr:    from /usr/local/bin/bundle:23:in `<main>'

Is it a problem with my bundler? the path /usr/lib/ruby/bin/ does not exist
on my machine.

[image: image]
<https://cloud.githubusercontent.com/assets/2731594/5918139/4bbb64ba-a5f3-11e4-8c0b-1eb0ea5b605e.png>

—
Reply to this email directly or view it on GitHub
<https://github.com/concerto/concerto/issues/1174>.

snoop168:
```
[Tue Jan 27 08:16:42.859079 2015] [mpm_prefork:notice] [pid 3587] AH00169: caught SIGTERM, shutting down
[...]

PR #1178, opened 2015-02-03
remove unnecessary block for cancan rules

simplysoft:
fixes #1177 and #1174 

(The user rules themselves are meaningless because some lines above there is a rule that allows anybody to read all users, but that's another topic)


augustf:
@bamnet Thoughts on this?

bamnet:
LGTM!

PR #1192, opened 2015-03-17
Removing pre-1.9 hashrockets

augustf:
I think I've successfully removed all the hashrockets from the codebase and things are testing well.


zr2d2:
I disagree. The mysql gem needs to stay in the lock file. Moreover, why do syntax changes require alterations to gems?

zr2d2:
Also, there are circumstances where hashrockets still need to be used
* quoted symbols (:'the.word' => 'bird')
* invalid labels (:$foo => 'bar')
* non-symbolic keys

augustf:
'Twas not my intention to commit my local Gemfiles and I'm not even sure why they're missing mysql (as I have that installed). I've overwritten them with the ones from master.

augustf:
@zr2d2 On the uses of hashrockets, I think that in our codebase, only the first might be present. But I haven't myself seen any of this. Please feel free to drop in any commits along those lines.

Obviously, this is just a PR, so if we decide no to push the envelope with syntax, that's just fine too.

Issue #1156, opened 2014-12-21
Devise 3.4.x upgrade breaks on Travis.


augustf:
Upgrading Devise to 3.4.x requires that a secret_key config setting be defined in Devise. In a normal deployment, this can be solved by setting it to ConcertoConfig[:secret_token]. In Travis, however, the Devise initializer (even when placed after the secret token initializer) fails to find a ConcertoConfig table.

My sense is that Travis is not running the initializers in the usual order or is somehow preloading them, but it's hard to say.

mfrederickson:
It does look like that table does not exist when running under the test environment the first time.  When I run tests manually, locally, the first time it bombs with a no secret key base, the second time it works fine because the table exists.  Since the test db is recreated each iteration, why not allow it to set it and not store it to the table if the table doesn't exist yet?

augustf:
But I guess what I'm not following is why (since table creating is just about the first initializer) it doesnt exist yet. And you'll find that in many of the initializers, we have to kludge that issue with the config table not existing. I've yet to find out why.

mfrederickson:
Because of this:  https://github.com/concerto/concerto/blob/master/config/initializers/01-db_create_load_or_migrate.rb#L6

mfrederickson:
When I comment that out, testing works fine the first time.

augustf:
Ahh yes-I'd forgotten about that other little kludge. I believe we did that because Travis doesn't like auto-creating tables.

mfrederickson:
Does it still?  That was 2 years ago (#280).  Maybe we could detect Travis also on that line? http://stackoverflow.com/questions/12983137/how-do-detect-if-travis-ci-or-not

augustf:
I have no idea. That article suggests its in an environment variable. Testing anything like this is always a bit weird b/c we can't replicate that unique environment. I've no idea if any code we'd write was working until Travis perhaps failed on it.

zr2d2:
If you submit the change as a pr, you can find it if Travis likes it

PR #1200, opened 2015-03-30
add update and delete template tests 

mfrederickson:
Not sure how the rest of you have your environment set up, but I had to add a `SECRET_TOKEN: entry` to my concerto.yml for my tests to run locally.
```
rake test test/functional/templates_controller_test.rb 
```


augustf:
@mfrederickson Yes-and I don't know why. In theory, https://github.com/concerto/concerto/blob/master/config/initializers/07-secret_token.rb handles this, but for some reason isn't.

augustf:
In fact, I mention the secret_token issue in #1156

PR #1201, opened 2015-04-01
use __FILE__ for logging

mfrederickson:
Added the logging to those where it was missing, and also swapped devise and secret token initializers since the newer devise will need the token.

### DIFF
--- a/config/initializers/00-inflections.rb
+++ b/config/initializers/00-inflections.rb
@@ -1,5 +1,7 @@
 # Be sure to restart your server when you modify this file.
 
+Rails.logger.debug "Starting #{File.basename(__FILE__)} at #{Time.now.to_s}"
+
 # Add new inflection rules using the following format
 # (all these examples are active by default):
 ActiveSupport::Inflector.inflections do |inflect|
@@ -14,3 +16,5 @@ end
 # ActiveSupport::Inflector.inflections(:en) do |inflect|
 #   inflect.acronym 'RESTful'
 # end
+
+Rails.logger.debug "Completed #{File.basename(__FILE__)} at #{Time.now.to_s}"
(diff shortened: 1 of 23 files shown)


augustf:
@mfrederickson So you're thinking the order change will overcome the issue we were discussing in the Devise upgrade ticket?

mfrederickson:
No.  That's something else.  Devise 3.latest needs a secret key token, we're currently on 3.0.4 I think.  I'm going to look into that other problem later today-- it could be that it wont be a problem with the latest devise (or maybe easier to find a solution for).

zr2d2:
I like the interpolation of  `File.basename`. Now that all the messages are the same, could the initializer be passed as a block to a logging method to dry this up?

mfrederickson:
I'm not sure Zach, I looked for a before/after that would fire for each initializer, but it looks like only one fires before all of the them and only one after.

augustf:
@mfrederickson I think we can, in the new Devise config, just use our existing secret token for devise by setting it in the devise initializer, assuming it's in scope at that point.

mfrederickson:
@augustf it will be since we just swapped their order.

zr2d2:
just because it comes before doesn't mean it will be in scope. Ruby has
[scope gates](http://hopsoft.github.io/blog/ruby-metaprogramming-idioms/),
so scoping behaves differently than in other languages


Zach Rowe

augustf:
Luckily, I don't think anything that esoteric will come into play. The token is/can be stored in the DB or an environment variable.

zr2d2:
I would argue that knowing how ruby works is important when writing in
ruby, it is for your edification


Zach Rowe

augustf:
"Questions are a burden to others; answers a prison for oneself." --The Prisoner

mfrederickson:
Thanks for the edification and the banter-- both were enjoyable.

Issue #1157, opened 2014-12-21
Gemfile database adapter detection fails on new Rails/bundler


augustf:
Upon upgrading to the latest stable Rails, I have found that our mechanism for conditionally including mysql and postgres gems in the Gemfile is not working as expected. To reproduce this, simple upgrade Rails/bundler/et al and issue the command `bundle update mysql2`. Bundler will say there is no such gem.

augustf:
We're not going to be able to upgrade gems much further without hitting this bug, so I've designated it a blocker for future releases.

augustf:
@mfrederickson Are you seeing this issue anymore?

mfrederickson:
Yes, I see the same problem that you report in the original message above.

mfrederickson:
It is in the production group, does that matter? I tried in my development environment.

augustf:
Yeah-I don't think they're supposed to show up in development mode. This should be a production mode issue.

mfrederickson:
Yeah, I suppose. But I run some of my other projects in mysql2 in dev.  Anyway when I prefix it with RAILS_ENV=production I still get the problem.

mfrederickson:
I've seen a comment that says their fix was to remove the Gemfile.lock and try again.  I tried this and that solution worked for me.  This makes sense to me when you think about it-- that gem is only getting included in production and since the Gemfile.lock is produced by the developer and checked in, it's getting locked out (so to speak).  I might be grossly misunderstanding this.  If so please let me know.  I suppose I'd fix this by always including that gem regardless of environment.

augustf:
So it turns out that issues related to this have been breaking the build for some time. I never received the email, and so only found out now. I've tried to mend the Gemfile.lock with no success. I think the next step will be to move my last 2 commits out to a branch and revert them from master until some solution is found.

PR #1202, opened 2015-04-01
update for devise 3.4

mfrederickson:
bundle updated and slightly modified for devise 3.4. also changed the travis.yml and our db initializer so travis runs like we would (with our app running the migrations) and locally rake test works also (without running twice or having to manually create the db first).

-- previously --
Since I don't know how to get travis to check this commit/branch, I'm going to create this PR, but please don't merge it until I change this message.


zr2d2:
Long-term, you would add the branch name [here](https://github.com/concerto/concerto/blob/master/.travis.yml#L22) to get it to build on new commits. A PR is the way I would go for testing like this. And **some people** shouldn't be so trigger-happy

augustf:
This is how we do Science. Also, it's good to be a king.

mfrederickson:
@zr2d2 @augustf I'm going to change the devise restriction from ~>3.0.0 to just ~>3 and run a `bundle update` and then add the Gemfile and Gemfile.lock here so it actually comes up to devise 3.4, ok?

augustf:
Sure! If we can do it without issues, I'd like to do a comprehensive bundle update.

mfrederickson:
There were no problems when I did it locally, your majesty.

augustf:
Haha indeed good esquire! I got burned by #1157 last I tried it.

zr2d2:
You could technically use~> 3.0, but sure. @augustf, that would only work
if it's a dependency problem, but i thought the problem was with the tables
not being created

mfrederickson:
If the screens work and the player works with the new devise then I'm fine with going ahead.  I'm away from the computer for the next couple hours.

mfrederickson:
The latest concerto player image seems to work fine with this branch (with tokens).

PR #1207, opened 2015-04-10
Improve plugin install process

augustf:


augustf:
NoMethodError: undefined method `make_plugin_configs' for #<Class:0x00000004226380>
/home/travis/build/concerto/concerto/vendor/bundle/ruby/1.9.1/gems/activerecord-4.1.10/lib/active_record/dynamic_matchers.rb:26:in `method_missing'

PR #1208, opened 2015-04-10
Trying to fix a symptom of #1157

augustf:


mfrederickson:
Is this a problem with travis because the screen table had it's width removed and the fixtures are still trying to populate it?  That what it looks like to me.

augustf:
I'm not really sure what to make of Travis' output -- but that seems like a good move to me.

augustf:
This all seems to have been a mistake on our part in an earlier commit. We took a decision to remove width and height without realizing they're used in the screen model.

PR #1213, opened 2015-04-18
redo ce3ec3, don't upgrade gems willy nilly

zr2d2:


augustf:
Some people say gems are hard and scary...

Issue #1238, opened 2015-09-08
jobs:work: Octokit::TooManyRequests: GET https://api.github.com/users/concerto-addons/repos: 403 - API rate limit exceeded


salfter:
The script that updates RSS feeds keeps crashing.  It keeps trying to access GitHub, but runs up against a rate limit.  It says that authenticated requests get a higher limit, but AFAIK there's no information in the Concerto documentation on setting authentication parameters.

Here's the trace output:

```
** Invoke jobs:work (first_time)
** Invoke jobs:environment_options (first_time)
** Invoke environment (first_time)
** Execute environment
rake aborted!
Octokit::TooManyRequests: GET https://api.github.com/users/concerto-addons/repos: 403 - API rate limit exceeded for 24.120.210.226. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) // See: https://developer.github.com/v3/#rate-limiting
/var/www/concerto/vendor/bundle/ruby/1.9.1/gems/octokit-3.8.0/lib/octokit/response/raise_error.rb:16:in `on_complete'
/var/www/concerto/vendor/bundle/ruby/1.9.1/gems/faraday-0.9.1/lib/faraday/response.rb:9:in `block in call'
/var/www/concerto/vendor/bundle/ruby/1.9.1/gems/faraday-0.9.1/lib/faraday/response.rb:57:in `on_complete'
/var/www/concerto/vendor/bundle/ruby/1.9.1/gems/faraday-0.9.1/lib/faraday/response.rb:8:in `call'
/var/www/concerto/vendor/bundle/ruby/1.9.1/gems/faraday-0.9.1/lib/faraday/rack_builder.rb:139:in `build_response'
/var/www/concerto/vendor/bundle/ruby/1.9.1/gems/faraday-0.9.1/lib/faraday/connection.rb:377:in `run_request'
/var/www/concerto/vendor/bundle/ruby/1.9.1/gems/faraday-0.9.1/lib/faraday/connection.rb:140:in `get'
/var/www/concerto/vendor/bundle/ruby/1.9.1/gems/sawyer-0.6.0/lib/sawyer/agent.rb:94:in `call'
/var/www/concerto/vendor/bundle/ruby/1.9.1/gems/octokit-3.8.0/lib/octokit/client.rb:339:in `request'
/var/www/concerto/vendor/bundle/ruby/1.9.1/gems/octokit-3.8.0/lib/octokit/client.rb:200:in `paginate'
[...]

PR #1239, opened 2015-09-08
Partially address #1238 by not letting Rake stuff call the Github API

augustf:


augustf:
@mikldt What are your thoughts on this?

PR #1243, opened 2015-09-12
fix error when listing pages and there arent any

mfrederickson:


augustf:
This is a reasonable change, but I'm unsure on why there wouldn't be any pages. There's at least one included in every seed, and has been for quite a few versions.

mfrederickson:
Yeah, I had a page, but had to delete it to get that error.

Issue #1250, opened 2015-09-18
Revisit i18n


augustf:
We need to finalize our work on i18n in Concerto. I haven't gotten to look over the views or helpers yet, but I believe they're in much better shape than the controllers were. My goals here were to minimize duplication of strings and make sure a translator will hit all user-viewed strings when they go through translating.

Before merging this, we should do the following:
- Go over all of our i18n work in MVC areas
- Optimize the verbiage in the YML files
- Check on non-string localization (and remove symbols and measurements we don't use in the app)
- Sort out #1143 (which deals with date/time formats)
- Finalize our technology decision (YML vs gettext et al) <sup>1</sup>
- Decide on one file or many for i18n (the plethora of view yml's, which once seemed like a nice structure, now seems maybe a bit odd to me)
- Settle on a translation service (there's already a German translation done in Locale)<sup>2</sup><sup>3</sup>

By stabilizing our i18n work, we'll be able to commit to getting folks to seriously work on translating things. There's only some many times we can reasonably put people through this, after all.  As it is, my controller work has necessitated a fresh translation.

<sup>1</sup>https://lingohub.com/blogs/2013/08/ruby-gettext-internationalization-tutorial-fest-gettext-gem/

<sup>2</sup>https://translation.io/

<sup>3</sup>https://www.localeapp.com/projects/7725

simplysoft:
Hi

I'd like to reference #929 here. Some aspects have already been discussed there. 
The most complete german (and french, dutch and hebrew) is available at https://crowdin.com/project/concerto however the strings were not updated for some months. 

Our take on some of the bullet points issues are:
- File Structure: use component (view / controller) spitted yml for english language and one big file for the translations (as it is right now)
- Translation Service: CrowdIn is a professional grade tool and free for open source projects. It provides e.g. a vital translation memory that other basic amateur / semi-professional tools don't have. It also allows for review of the strings, so you can allow anybody to translate concerto. We have already invited you as an project owner.

augustf:
@simplysoft So you wouldn't say there's any benefit in combining the English files as well?

PR #1251, opened 2015-09-18
Revisiting i18n

augustf:
This branch represents the beginning of finalizing our i18n work. The overall work on this is being tracked in #1250

### DIFF
--- a/app/controllers/concerto_plugins_controller.rb
+++ b/app/controllers/concerto_plugins_controller.rb
@@ -49,7 +49,7 @@ class ConcertoPluginsController < ApplicationController
         restarted = restart_webserver()
       end
       if restarted
-        flash[:notice] = t(:plugin_created)
+        flash[:notice] = t(:was_created, name: @concerto_plugin.name, theobj: t(:plugin))
       end
       redirect_to concerto_plugins_path
     else
@@ -68,7 +68,7 @@ class ConcertoPluginsController < ApplicationController
         flash[:notice] = t :plugin_updated_frozen_env
       else
         write_Gemfile
-        flash[:notice] = t(:plugin_updated)
+        flash[:notice] = t(:was_updated, name: @concerto_plugin.name, theobj: t(:plugin))
       end
       redirect_to concerto_plugins_path
     else
@@ -99,7 +99,7 @@ class ConcertoPluginsController < ApplicationController
       restarted = restart_webserver()
     end
     if restarted
-      flash[:notice] = t(:plugin_removed)
+      flash[:notice] = t(:was_deleted, name: @concerto_plugin.name, theobj: t(:plugin))
     end
     redirect_to concerto_plugins_path
   end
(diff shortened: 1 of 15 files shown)


simplysoft:
A lot of intended duplication for translation was removed and you are now again trying to generate human language, which tends to be easy for english, but can be hard with e.g. Japanese or impossible with even more obscure languages. Turns out that you can't easily generate sentences for all languages using a fixed structure. This is why human translators are needed.

See also our comments at https://github.com/concerto/concerto/issues/1135#issuecomment-68591222

app/controllers/concerto_plugins_controller.rb
```diff
@@ -49,7 +49,7 @@ def create
         restarted = restart_webserver()
       end
       if restarted
-        flash[:notice] = t(:plugin_created)
+        flash[:notice] = t(:was_created, name: @concerto_plugin.name, theobj: t(:plugin))
```
simplysoft, line 52: Coders love to reduce duplication, but this does not work for natural languages. It looks like this duplication is a lot of unnecessary work for translators, but it makes their life actually easier. They can e.g. use Translation memory to speed up the process and still have the ability to respect some language specifics when translating. And it is much easier to translate a real sentence than something like `The %{name} %{theobj} was successfully created.` where the translator needs to account for all future possible words that might be filled in there. 

And this simple string, can't be translated to a pretty similar language to english, which happens to be German. There we would have to distinguish between male and female `%{name}`and use different articles and can't just use `The`. So in german it should be `%{article} %{name} %{theobj} was successfully created.`. 

When you are trying to translate this into even more different languages, that have complexity different grammatical structures and concepts, it is impossible to correctly translate this.  
You will end up with `%{article} %{name} %{theobj} %{temporalverb} %{adjective} %{verb}.`

augustf:
@simplysoft I'm afraid you're right. I thought I could save this scheme by allowing essentially recursive translations, but that won't work. We presently have a situation where a substantial minority of our strings are untranslated as they've not been entered in a YML file, but at present, I don't have a solution.

augustf:
@simplysoft Would many of the lines we have such as  ```are_you_sure_delete_model_key: Are you sure you want to delete %{model} "%{key}"?``` also be problematic with your translations?

simplysoft:
@augustf Yes, this also would apply to ```are_you_sure_delete_model_key: Are you sure you want to delete %{model} "%{key}"?```. For german it works, but I bet there is some language out there where it would not work. 
One middle course could be to still use those composing yml strings but allowing to overwrite a default translation with specific ones. We have already used this e.g. here:

https://github.com/concerto/concerto/blob/master/app/views/layouts/_errors.html.erb
https://github.com/concerto/concerto/blob/master/config/locales/de.yml#L653-L692

This still allows to only add one generic translation (for languages where this works), but gives the flexibility to add special translations for each special case. 
(BTW: Writing out all possible cases with their full translation key is for better tool support, refactoring, translation support etc.)

PR #1259, opened 2015-10-15
Frontend cache key update

mfrederickson:
partial fix for concerto/concerto-frontend#50.  Includes media change detection and fixes a bug where passed arrays or associations were not having their updated_at property checked.


mfrederickson:
fyi, the Travis failure is due to OctoKit overusage...

PR #1275, opened 2015-11-02
handle null select_values as an empty selection list

mfrederickson:


augustf:
This came up as a result of having no values for certain fields like mail server type present after an upgrade. I guess null-checking is ok, but this really bypasses the issue itself. If the select values for a select field aren't in the database, we have a fatal problem. I'm not sure we should hide this from users/devs.

mfrederickson:
This wasn't intended to hide the problem.  right now when it occurs, the application crashes and you have to sleuth around to figure out why it crashed-- this just allows it to not break, so that you can see that the selection list is empty-- when it shouldn't be.  Throwing an alert when this happens would be good too.

augustf:
Right-so I don't have a problem with this - but let's try to find out what's going on.

PR #1281, opened 2015-11-03
closes #1280 show templates custom css if any

### LINKED ISSUE #1280: show css content next to image when previewing a template
Currently you can see the background image for a template, but you cannot see the CSS that was uploaded for that template.  This will provide a way for people to see it so they can use it if they want when creating a new template, or tweaking an existing one.

Valid request?

mfrederickson:
If you have supplied a template with some custom css, you can now see what that custom css content is in case you want to tweak it or copy it to another template.

image

### DIFF
--- a/app/controllers/templates_controller.rb
+++ b/app/controllers/templates_controller.rb
@@ -42,6 +42,11 @@ class TemplatesController < ApplicationController
     @template = Template.find(params[:id])
     auth!
     # the form contains two bogus fields used for file uploads -- :template_css, :template_image
+
+    css_media = @template.media.where({key: 'css'})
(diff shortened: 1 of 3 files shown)


augustf:
Would it be a good idea to put this in line with the rest of the fields and have it expand (since it will likely be seldom-viewed)? Doing it this way may seems to clutter things a bit and may make things harder for smaller devices and future responsive layouts.

mfrederickson:
@augustf it uses the row-fluid class and is responsive.  If there is no custom css, it won't show.  If there is it will, and if there's not enough room, it will show under the sample image.

PR #1287, opened 2015-12-04
fixes #1286 non-simple cors headers

### LINKED ISSUE #1286: Refused to get unsafe header "etag"
While working on #1284 I came across this error in the console when viewing the frontend.  It was also occurring on the X-Concerto-Frontend-Setup-Key header.

mfrederickson:
Would like some guidance on making this DRY.  Since it's only two lines of code, I wasn't sure what the best approach would be.   Works fine as is, and I'm ok with it.  However, same conditional on those two lines of code, so they could be consolidated into an `if...end`, and also two different files and wasn't sure if moving it to the application controller would be too much.  Like I said, small change, might not matter...


mfrederickson:
Looks like this should maybe be moved to the allow_cors method in the application controller.

mfrederickson:
No this commit is not needed.

PR #1290, opened 2015-12-09
add task to copy a concerto template

mfrederickson:
Now you can run a rake task to copy a template!
```
rake template:copy["BlueSwoosh","PurpleRain"]
```
Very important to *not* have a space after the comma in the parameters or rake will get irritated.
```
rake template:copy[source_name,destination_name]  # copy a concerto template
```


augustf:
@mfrederickson How is the _copy partial being used? It doesn't seem to get referenced or used anywhere in the PR.

mfrederickson:
It's called via the public activity view dynamically because the key is template.copy

PR #1293, opened 2015-12-11
additional public activity for membership

### LINKED ISSUE #1277: Promotion to group leader not logged to activity log

mfrederickson:
closes #1277 
Adds more pa logging by setting the pa key to the custom update action in the membership controller.

### DIFF
--- a/app/views/groups/manage_members.html.erb
+++ b/app/views/groups/manage_members.html.erb
@@ -92,7 +92,7 @@
               <thead>
                 <tr>
                   <th><%= User.human_attribute_name(:name) %></th>
-                  <th><%= t('.request_made_at') %></th>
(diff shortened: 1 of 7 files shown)


app/views/groups/manage_members.html.erb
```diff
@@ -92,7 +92,7 @@
               <thead>
                 <tr>
                   <th><%= User.human_attribute_name(:name) %></th>
-                  <th><%= t('.request_made_at') %></th>
+                  <th><%= I18n.t('.request_made_at') %></th>
```
mfrederickson, line 95: I know-- this shouldn't be needed, but it wouldn't work without it. And even wierder, it was resolving the path to the root... hence the entry in the en.yml.

Issue #1279, opened 2015-11-03
Previously Denied Members toggle


mfrederickson:
Who knew?  If you click on that header, it will show you a list of who was previously denied membership.  This should be more apparent.  In addition, the column headers for this table need translated and proper-cased

PR #1298, opened 2015-12-16
Fix plus/minus icons for #1279

augustf:


mfrederickson:
Ah... I must have missed those when I updated the font-awesome a while back... sorry.

mfrederickson:
Can you fix the headers?
image

augustf:
Yes-and actually I have that in with a separate i18n issue I ran into on the way.

Issue #1285, opened 2015-11-30
ActionView::Template::Error (undefined method `children' for nil:NilClass)


mfrederickson:
From groups:
After upgrading from 2.2.7 to 2.3.1, I am getting a 500 error on the actual server gui that says "Something went wrong. Rest assured, you have been assigned 0% of the blame."

So far this only happens in 1 feed, but I will be testing it in others.

Installed using apt-get on ubuntu.

Here is the error I get in /var/log/apache2/error.log:
```
App 12516 stderr: ** [Airbrake] Success: Net::HTTPOK
App 12516 stderr:
App 12516 stderr: ActionView::Template::Error (undefined method `children' for nil:NilClass):
App 12516 stderr:     1: <div class="tile<% unless submission.content.children.empty? %> tile-stack<% end %>" data-submission-id="<%= submission.id %>">
App 12516 stderr:     2:   <%= link_to [submission.feed, submission], :class => "tile-content" do %>
App 12516 stderr:     3:     <%= render_content(submission.content, {:type => 'tile'}) %>
App 12516 stderr:     4:
App 12516 stderr:   app/views/submissions/_submission_tile.html.erb:1:in `_app_views_submissions__submission_tile_html_erb__3906644533564743768_44304860'
App 12516 stderr:   app/views/submissions/_index_body.html.erb:186:in `block in _app_views_submissions__index_body_html_erb__3492203619809220758_46670740'
App 12516 stderr:   app/views/submissions/_index_body.html.erb:185:in `each'
App 12516 stderr:   app/views/submissions/_index_body.html.erb:185:in `_app_views_submissions__index_body_html_erb__3492203619809220758_46670740'
App 12516 stderr:   app/views/submissions/index.html.erb:6:in `_app_views_submissions_index_html_erb___1888457661993994339_45379140'
App 12516 stderr:   app/controllers/application_controller.rb:112:in `set_time_zone'
App 12516 stderr:
App 12516 stderr:
App 12516 stderr: Processing by ErrorsController#render_error as HTML
App 12516 stderr:   Parameters: {"state"=>"denied", "feed_id"=>"2"}
App 12516 stderr: Completed 200 OK in 57ms (Views: 0.6ms | ActiveRecord: 15.7ms)
App 12516 stderr:   Rendered errors/error_generic.html.erb within layouts/no-topmenu (2.5ms)
App 12516 stderr:   Rendered elements/_head.html.erb (0.2ms)
[...]

PR #1299, opened 2015-12-16
Check for submission with nil content for #1285

augustf:


augustf:
@mfrederickson Does this look reasonable? I'm actually not 100% clear where that first div ends.

app/views/submissions/_submission_tile.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="tile<% unless submission.content.children.empty? %> tile-stack<% end %>" data-submission-id="<%= submission.id %>">
+<div class="tile<% if !submission.content.nil? && !submission.content.children.empty? %> tile-stack<% end %>" data-submission-id="<%= submission.id %>">
   <%= link_to [submission.feed, submission], class: "tile-content" do %>
     <%= render_content(submission.content, {type: 'tile'}) %>
```
mfrederickson: What about here if submission.content is nil?

augustf: Well, if it is nil, the block won't be displayed. That's what we want, right?

mfrederickson: You're only affecting the class so the "stacked" icon does not appear if there are no children (or no content).  I'd think you'd want to make that render conditional also (if the content was not nil), at the least, or it will blow up there next.

mfrederickson: But maybe the partial checks for that already?

augustf:
@mfrederickson So in 762e561c9cf7dafda1cec24b8003439fbbce65b9 I went a little further. How about that?

mfrederickson:
From a cursory glance, it looks good.  (Got some screens I need to get running...)

augustf:
Resolved via ec52bce22bc579c0964883cb8e494bc33307163f

PR #1310, opened 2015-12-24
Migrate to Rails 4.2

gbprz:


gbprz:
Seems like I overlooked some more issues with Ruby 1.9.3. I'll reattempt this when those are sorted out.

gbprz:
Let me know if there are any thoughts on #1311

Gemfile
```diff
@@ -36,35 +36,37 @@ gem 'jquery-timepicker-rails'
...
+
 #RMagick is used for image resizing and processing
-gem 'rmagick', :require => 'rmagick', :platforms => :ruby
+gem 'rmagick', require: 'rmagick', platforms: :ruby
 
 # Attachable does all the file work.
 gem 'attachable'
 
-gem 'devise', '~> 3'
+gem 'devise', '~> 3.5'
```
zr2d2: do we want minor versions to float, or just tiny? ~> 3.5 means <4, whereas ~> 3.5.3 would mean <3.6

Gemfile
```diff
@@ -77,13 +79,16 @@ gem 'google-analytics-turbolinks', '~> 0.0.4'
 
 # I18n Tasks
 group :development do
-  gem 'i18n-tasks', '~> 0.7.10'
+  gem 'i18n-tasks', '0.8.7'
```
zr2d2: could we bump to ~> 0.9.2?

Gemfile
```diff
@@ -77,13 +79,16 @@ gem 'google-analytics-turbolinks', '~> 0.0.4'
...
   gem 'slop', '~> 3.6.0' # Required due to https://github.com/glebm/i18n-tasks/issues/118
 end
 
 gem 'i18n-js', '>= 3.0.0.rc8', '< 3.1.0'
 
 #Github API
-gem 'octokit'
+gem 'octokit', '~>4.0'
```
zr2d2: same as devise, do we want minor versions to float?

config/initializers/13-airbrake.rb
```diff
@@ -8,22 +8,13 @@
...
-          return nil
-        end
-      end
-      #config.async = (RUBY_VERSION.to_f > 1.8)
-      config.user_attributes = []
-      config.host = 'errors.concerto-signage.org'
-      config.port = 80
-      config.secure = config.port == 443
-      config.environment_name = Concerto::VERSION::STRING
+      config.project_key = if ConcertoConfig[:send_errors] then '34e36775df3e89293c59efeba36f6c8f' else '' end
```
zr2d2: I think this might be cleaner:

``` ruby
config.project_key = '34e36775df3e89293c59efeba36f6c8f' if ConcertoConfig[:send_errors]
config.project_key ||= ''
```

thoughts?

PR #1318, opened 2016-02-02
show subscribed content for each screen

mfrederickson:
Outside of normal viewblock-content so it is not confined by the existance of a sidebar.
No pagination.
Merge if you want, close if you don't.


augustf:
I don't think this is where we want to go with the screen#show view right now. In any reasonably large installation, this could be a bit odd and might give people the idea that the content will be in a order that might not be the case.

mfrederickson:
just an fyi-- this has been accomplished through a plugin-- https://github.com/mfrederickson/concerto-screencontent

Issue #1313, opened 2016-01-14
image size warning


mfrederickson:
It would be a good idea to display the image size somewhere under the preview when adding an image, and give a warning if it is larger than a predetermined size...

augustf:
@mfrederickson Do you have any notion of sizes where we're starting to see problems. And is this a matter of MB of size vs pixel size?

mfrederickson:
I think it was actually pixel density/depth/resolution or whatever you want to call it that was causing problems for Matthew.  But the large MB size was also causing slow display times (sometimes the picture did not have enough time to download and display before the screen called for the next piece of content).

mfrederickson:
Closing since  #1320 already shows the size info, and the criteria of the warning would vary too much based on hardware/network/user needs.

PR #1320, opened 2016-02-03
display image size info

mfrederickson:
It's a start towards #1313.  But we also need to display this somewhere when browsing the feed content...


augustf:
@mfrederickson To make this deal with the silent failures we've had reported, there should be a configuration parameter for the max size and some validation to enforce it.

PR #1330, opened 2016-03-20
Make frontend fields stateless to resolve repeating content issues.

### LINKED ISSUE #1329: Repeating content on frontend
We've had quite a few reports on the mailing list about content in the rotation repeating. The post ( https://groups.google.com/forum/?utm_source=digest&utm_medium=email#!topic/concerto-digital-signage/xtWKhTK4k9o) and some follow-up emails have been able to confirm this is happening.

mfrederickson says "probably something in the frontend or the caching.  The next step to take I think is to write a simple ruby script that pulls content down from the server (just for the graphics field) to see if it repeats from a simple client. One of the emails in that chain lists how to pull back the list of content in the order it should be coming in. (rails console code using the base shuffle)"

mikldt:
This eliminates the concept of a store for timeline data, which was just going to the client anyway when cookie store was enabled. Instead, the entire timeline worth of content will be sent to the client. This timeline is not reliant on getting a cookie back which could have been overwritten in the client by a simultaneous request.

This will also have the side effects of reducing overall requests to the server, reducing the size of cookies that need to be sent to the server with each request, and reducing the overall number of queries from each screen for a given period of time.

If a reliable store for this data is found, the store concept could be reinstated. However the only reason for this would seem to be if the client could not store the whole timeline, which seems unlikely.

This seems to be effective in testing on my local box. If other folks are OK with the approach (@augustf @mfrederickson @gbprz) , we should test and merge. This would close #1329.

### DIFF
--- a/app/controllers/frontend/contents_controller.rb
+++ b/app/controllers/frontend/contents_controller.rb
@@ -22,19 +22,14 @@ class Frontend::ContentsController < ApplicationController
 
     shuffle_config = FieldConfig.get(@screen, @field, 'shuffler') || DEFAULT_SHUFFLE
     shuffler_klass = FrontendContentOrder.load_shuffler(shuffle_config)
-    session_key = "frontend_#{@screen.id}_#{@screen.template.id}_#{@field.id}_#{shuffler_klass}".to_sym
     shuffler = nil
-    count = 1
-
-    count = 20 if FieldConfig.get(@screen, @field, 'marquee') == '1'
 
     run_callbacks :index do # Run plugin hooks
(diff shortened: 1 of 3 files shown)


mfrederickson:
If a screens template is changed on the backend while a screen is being displayed on a frontend does the frontend still immediately refresh and show the new template?

gbprz:
I just pulled the branch to give it a try and everything works great! Changing the template, adding/editing field configs, and changing the timezone (all tested separately) worked well. Contents load just as intended too.

augustf:
This is looking good on the demo instance

PR #1362, opened 2016-12-01
Add content fullcalendar and add content list button

neeev:


augustf:
@neeev Currently, your PR is failing our integration tests. In particular, the ones that deal with content URL's and content mixing with feeds. While we examine this PR, can you tell us a bit about your goals with it?

neeev:
@augustf I added "fullcalendar" that permit to see content planning and I also added a link to the content search page.

augustf:
So everyone can see what's being proposed here: 
screen shot 2016-12-03 at 8 57 11 pm

neeev:
I think it's a major part of this webapp. I use concerto and when i create content non associated to a feed I can't find my content with the actual menu.

mfrederickson:
Translation is missing:
image

Thanks @neeev.  I do like the calendar, and I like the functionality of a search-- would it be feasible to add a search by name/title?  Are the select lists filtered to just what the user is allowed to see? Something about the UI just doesn't sit well with me at the moment... though I'd be hard pressed to say just what it was...

mfrederickson:
The content thumbnails in the search results also seem to be missing their "type" icons in the lower right corners.

neeev:
@mfrederickson I fix the translation problem

mfrederickson:
Yes, thank you.  @augustf is the one that merges external PRs.  What about those icons?
image

Issue #1345, opened 2016-05-22
Automatically precompile assets when installing / upgrading Concerto plugins


gbprz:
All of the code exists to run a rake task for precompiling assets:
[Helper method to invoke Rake Task](https://github.com/concerto/concerto/blob/f5cd0a1d6709f02bf6a96ce86b9d496aaeec3566/app/controllers/application_controller.rb#L133)
[Install a new Gem calls helper](https://github.com/concerto/concerto/blob/f5cd0a1d6709f02bf6a96ce86b9d496aaeec3566/app/controllers/concerto_plugins_controller.rb#L49)
[Update a Gem calls helper](https://github.com/concerto/concerto/blob/f5cd0a1d6709f02bf6a96ce86b9d496aaeec3566/app/controllers/concerto_plugins_controller.rb#L111)

But I don't think the actual rake task exists (let me know if I'm blind and can't find it), so users are left to manually precompile assets for plugins including the frontend and weather. It might save some trouble if this is added in since people run into black screens or missing weather icons often.

augustf:
https://github.com/concerto/concerto/pull/1376

PR #1376, opened 2017-02-17
plugin update feedback and alternate assets precompile

mfrederickson:
Addresses #1345 when upgrading a plugin... handles gem update async and provides more detailed feedback when bundle update fails or the assets precompile fails.  @augustf I changed the rake task to run from a system call so we could capture the output and success result-- don't know if that's a problem or not.  Seems to work fine for me in development, but a better test would be on a deployed environment.  

Please test this in your environments before you merge it.  Thanks.

### DIFF
--- a/app/controllers/concerto_plugins_controller.rb
+++ b/app/controllers/concerto_plugins_controller.rb
@@ -106,14 +106,22 @@ class ConcertoPluginsController < ApplicationController
   end
 
   def update_gem
+    results = {}
     plugin = ConcertoPlugin.find(params[:id])
-    system('bundle update', plugin.gem_name)
-    rake_precompile()
-    restarted = restart_webserver()
-    if restarted
-      flash[:notice] = t(:plugin_updated)
+    results[:bundle_output] = `bundle update #{plugin.gem_name}`
+    results[:bundle_success] = $?.success?
+
+    if results[:bundle_success]
+      results[:rake_output] = `bundle exec rake assets:precompile`#rake_precompile()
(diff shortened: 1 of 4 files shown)


app/controllers/concerto_plugins_controller.rb
```diff
@@ -106,14 +106,22 @@ def destroy
...
 
   def update_gem
+    results = {}
     plugin = ConcertoPlugin.find(params[:id])
-    system('bundle update', plugin.gem_name)
-    rake_precompile()
-    restarted = restart_webserver()
-    if restarted
-      flash[:notice] = t(:plugin_updated)
+    results[:bundle_output] = `bundle update #{plugin.gem_name}`
```
mfrederickson: @augustf should any of the parameters used in the bundle install from the boot.rb line 57 be included here?

augustf: @mfrederickson That command is a bit funny - it tends not to do what you expect: http://stackoverflow.com/questions/11263150/update-just-one-gem-with-bundler . We should use bundle update --source gem-name

config/routes.rb
```diff
@@ -17,8 +17,8 @@ def matches?(request)
   end
 
   resources :concerto_plugins do
-    collection do
-      post :update_gem
+    member do
+      get :update_gem
```
augustf, line 21: @mfrederickson Using a get is probably not ideal here. We're performing a system operation that has a bunch of substantial side effects (https://www.w3.org/2001/tag/doc/whenToUseGet.html).

PR #1377, opened 2017-02-17
content coverter better pdfs using pdftoppm

### LINKED ISSUE #1324: PDF conversion quality (again)
It seems the quality of the conversion is still not as good as it should be.  Suggest testing one of these methods instead of relying solely upon docsplit when going from PDF to JPG/PNG.  `-trim` would also benefit maximizing the content.

http://stackoverflow.com/questions/6605006/convert-pdf-to-image-with-high-resolution

For past issue regarding PDF quality see #973.

mfrederickson:
closes #1324

### DIFF
--- a/lib/concerto/content_converter.rb
+++ b/lib/concerto/content_converter.rb
@@ -72,7 +72,7 @@ module Concerto
 
         # process it with docsplit via the api
         begin
-          if original_filepath.end_with? ".pdf"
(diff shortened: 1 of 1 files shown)


augustf:
Added task to debian repo to get this in the control file: https://github.com/concerto/concerto-debian/issues/64

augustf:
@mfrederickson Are we falling back to something if existing users don't have pdftoppm installed?

mfrederickson:
Yes, docsplit.

PR #1379, opened 2017-02-17
edit ticker content

### LINKED ISSUE #1349: Edit Ticker Content
Is it possible to edit the text content of a ticker entry? Currently when I go to edit the text is static and all that can be changed is Start & End Times, the feed(s) and the duration... We often have a ticker entry that's always running but the content of it changes slightly almost on a daily basis. Of course the user can just deny the old entry and create a new one in its place but just thought it was a little more intuitive for the user to edit and make slight changes rather than copy->paste to new entry then make changes and of course don't forget to deny the old one or you will have both running...

mfrederickson:
closes #1349 by allowing content types to have custom edit forms.
(also renamed a partial to include the .erb extension)


augustf:
@mfrederickson I'm unable to reconcile the changes to _edit_form.html in this commit with the past PR of https://github.com/concerto/concerto/commit/47a8c67dd419fc14e51d95ae0b7bd4221a244aab and I'm unable to access this branch via my command line, so I can't even tell you the exact reason. Perhaps I merged them in the wrong order?

mfrederickson:
That file got renamed to have the .erb extension.  I can look at it I get home.

PR #1384, opened 2017-02-22
lock turbolinks at 2.5.3

mfrederickson:
This is where it was before the latest bundle update.


mfrederickson:
@augustf this resolves that problem that I emailed you about.

PR #1385, opened 2017-02-23
add a view hook under the screen details

mfrederickson:
I'd like to be able to display additional information via new plugins under the screen detail area.

### DIFF
--- a/app/views/screens/show.html.erb
+++ b/app/views/screens/show.html.erb
@@ -1,6 +1,7 @@
-<% sidebar = ConcertoPlugin.render_view_hook self, :screen_details %>
-
 <%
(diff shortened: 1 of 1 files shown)


augustf:
@mfrederickson What's the use case for this look like. Do we have plugin(s) using this?

mfrederickson:
@augustf the use case is for any plugin to be able to display additional information on the screen details page in an area underneath (not just constrained to a sidebar like we already have). 

Yes, I have a plugin that I just developed to show active content assigned to the screen, and have some future tweaks in mind to make it more useful.

I believe this has minimal impact to Concerto-- it's just an extension point.

PR #1387, opened 2017-03-01
closes #1383 update to turbolinks v5

### LINKED ISSUE #1383: update to turbolinks v5
We will probably have to update several js files because `page:load` has been replaced with something else.

mfrederickson:
This may also require concerto plugins to be updated so they don't break.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,7 +34,7 @@ gem 'uglifier', '~> 2.7.2'
 gem 'bootstrap-datepicker-rails'
 gem 'jquery-rails'
 gem 'jquery-timepicker-rails'
-gem 'turbolinks', '~>2.5.3'
+gem 'turbolinks', '~> 5.0'
 gem 'twitter-bootstrap-rails-confirm'
 
 gem 'responders', '~> 2.0'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -303,8 +303,9 @@ GEM
     thread_safe (0.3.5)
     tilt (1.4.1)
     tins (1.13.2)
-    turbolinks (2.5.3)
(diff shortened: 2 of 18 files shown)


augustf:
@mfrederickson Is this change backward compatible?

mfrederickson:
No, I don't believe so-- none of the older plugins will work with it-- that's why I had to change them as well.

augustf:
@mfrederickson Well what I meant is that if I update and push the plugins, and folks with older Concerto's install them, will it break?

mfrederickson:
Yes it will break

augustf:
@mfrederickson Then we'll probably need some mechanism to handle updates like this. Either include logic in the code change itself to avoid breakage or have a check for version number or something before installing it. We don't want to break the installations of folks who unwittingly upgrade a plugin.

mfrederickson:
I will look into making the plugins backwards compatible

mfrederickson:
@augustf I am unable to detect which version of turbolinks is running.  Do you have any ideas how we can make this backwards compatible?

augustf:
@mfrederickson My main thought would be to try to check by which method(s) it responds to. That is, if the new Turbolinks has new_turbolinks_method(), your script can test to see if that's true. Alternatively, we could expose the versions of installed Rubygems via an API of some sort. Failing all of that, we may need to find a way to restrict plugin upgrades in the panel so users don't break their installs.

augustf:
@mfrederickson Also, is a Turbolinks patch required for concerto-weather? I notice it alone of the addons didn't get one...

mfrederickson:
@augustf yes, it probably does need a patch.  Thanks.

PR #1388, opened 2017-03-01
garci66's priority shuffle (and other shuffle tests)

mfrederickson:


mfrederickson:
@augustf this includes some test cases for all the shufflers which would also be good to have in the repo.

mfrederickson:
@garci66 please update the field config wiki regarding this new shuffler, how it works and how it can be used.  Thanks.

garci66:
@mfrederickson wiki update is done. Thanks for reminding me!

Issue #1392, opened 2017-03-03
edit ticker needs security


mfrederickson:
#1379 didn't account for moderated content.  specifcally editing something that had already been moderated to contain content that a moderator would not want on their feed/screen.

The only people that should be able to edit actual content (outside of date ranges/feed subscriptions) are:
* system admin
* the owner if they are a moderator of all of the feeds where it has been submitted, or it hasn't been approved on any feed yet

mfrederickson:
According to the [abilities](https://github.com/concerto/concerto/blob/master/app/models/ability.rb#L107), only the user that created the content can edit it, and only if it has either not yet been approved, or has expired. And the system wide admin can edit it regardless of approval or expiration status.

mfrederickson:
Currently a user cannot update their own content if all of its submissions have been approved (unless it is expired).  I'm changing this to also allow them to edit their own content even if it has been approved on all of its submissions if they have the ability to moderate all of its submissions.

I'm also adding a new custom action called `:update_full_details` that will be true if they are the owner and if they have the ability to moderate all of its submissions.  This will satisfy the initial criteria in the first comment above.

mfrederickson:
I'm also changing what I considered to be another bug-- it was checking to see if the `content.is_approved? == false`, presumably to see if it had been approved yet.  But that method returns true only if it has been approved on all submissions.  So if 3 out of four submissions were approved, the user was still being allowed to edit the content.

mfrederickson:
I also changed the view to not allow editing if the piece of content is a child.  This is because parent items drive the details of the child items.

PR #1396, opened 2017-03-04
clean up/more secure editing of ticker and html_text content

mfrederickson:


mfrederickson:
Don't allow editing content of dynamic items such as rss feed items.

mfrederickson:
closes #1392

Issue #1449, opened 2017-06-30
Pulsating horizontal scrollbar on frontend screens


timparenti:
Manifests in current state of `master`.  The offscreen font-awesome icon introduced in 3820ea5cd93274f45a1580d35edc0d1435dd0e4b to address concerto/concerto-frontend#73 results in a horizontal scrollbar (which is quite visible in Firefox), since it's outside the `concerto-screen` which already has width and height of 100%.

Since the square icon rotates with period 2s, the scrollbar pulsates to accommodate its changing width with period 0.5s.

PR #1450, opened 2017-06-30
Hide scrollbars on frontend screens.

timparenti:
This resolves the pulsating scrollbar issue in #1449.


mfrederickson:
Thanks a bunch, @timparenti !

PR #1491, opened 2018-08-09
Solve SMTP server connection when credentials used

garci66:
ActionMailer user_name and password where being added into the Authentication key as an array (due to the additional commas). 
SMTP auth now working


garci66:
I can open a ticket if needed. Spent a few hours trobleshooting only to realise the SMTP was not being properly authenticated ( I am using a remote SMTP server with SASL auth)

mfrederickson:
My bad.  Good catch.  Thanks.

Issue #1496, opened 2018-11-27
Have anyone had any success with the install of Concerto2 on Ubuntu 18.04 or Debian 8 or 9


bossbarnes07:
Have attempted to install Concerto2 on both Ubuntu 18.04 and 18.10 both LTS and Debian both have problems with the install missing packages. Can anyone provide me a work around  beside the vm option. Thank you.

Kohlmeyer:
I also am having a really difficult time getting a running version of the ova or cloning Concerto 2 correctly. I would really like to see if there are solutions to this issue as well, thanks!

Ahmedjaseemiq:
me too I cant find any installation step for any Ubuntu or Debian version can work today, actually I start doubt  if there Concerto work!!, even Youtube I cant find.

mzocher:
I was able to get my server upgraded from Ubuntu LTS 16.04 to 18.04, but because it ships with Ruby 2.5 it took some tweaks:

- Before install, needed to explicitly install the older version of bundle.  Ran the command `gem install bundler -v 1.16.1`
- Needed to use the delayed_job gem version 4.14, as it adds Ruby 2.5 support (by switching from yaml_as to yaml_tag - see [https://github.com/collectiveidea/delayed_job/blob/master/CHANGELOG.md](url). Edited Gemfile.lock and changed the delayed_job version from 4.1.3 to 4.1.4
- Devise 3.5.10 has a bug that was fixed, but a new version was never replaced (see [https://github.com/plataformatec/devise/issues/4883].  You can point to the stable v 3 branch and get that new patch.  To do so, open Gemfile and replace `gem 'devise', '~> 3.5.5'` with `gem 'devise', :git=> "https://github.com/plataformatec/devise.git", :branch => "3-stable"`
-    Finally. ran dpkg --configure concerto-full to make sure everything was configured properly.

GrayMatter:
Came across this in the Concerto forum.  I haven't tried it but plan to give it a shot in the next week or two.

https://groups.google.com/d/msg/concerto-digital-signage/ywa9exSeA74/3JC6FsHdAQAJ

paulfurze:
Almost !!! .... but nope :(
Follwed this...
https://groups.google.com/forum/#!topic/concerto-digital-signage/ywa9exSeA74
but it eventually fails at...

`sudo dpkg --configure concerto-full`

Then the below happens.
I've no experience of ruby/rails/gems. Any pointers would be well received.

```
[...]

PR #1499, opened 2019-01-19
Minor adjustments for Ruby 2.5.x support

mzocher:
A few fixes for Ruby 2.5.x support, as discussed in https://github.com/concerto/concerto/issues/1496

PR #1500, opened 2019-01-19
Ruby 2.5.1 support

mzocher:
Per discussion on https://github.com/concerto/concerto/issues/1496 correcting some patch versions to work on ruby 2.4.5 and ruby 2.5.1.  Was compiling fine, but I am now adding the new versions into Travis CI test.


mzocher:
Ok, sorry for the noise everyone.  This one failed because I still had a test in progress from pull #1499.  I'll try again!

PR #1501, opened 2019-01-19
Ruby 2.5.1 Support

mzocher:
Per discussion on #1496 correcting some patch versions to work on ruby 2.4.5 and ruby 2.5.1. Was compiling fine, but I am now adding the new versions into Travis CI test.


mzocher:
Does this make sense in a different branch than master, so interested people can continue to modernize the gems involved?  2.4-dev perhaps?

PR #1556, opened 2021-03-05
Enable bundler caching for travis

Taher-Ghaleb:
Would be interested to know why caching hasn't been enabled in this repository. Thank you.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,8 @@ env:
 services:
   - mysql
 
+cache: bundler
+
 before_install:
   - sudo apt-get install -y gsfonts ghostscript libgs-dev poppler-utils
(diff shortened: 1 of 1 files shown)


Taher-Ghaleb:
Thanks for merging my pull request.

Any reasons why bundler caching wasn't enabled before, please? It was available since [December 2014](https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure).

Thank you.

bamnet:
No particularly good reason.  I don't remember our exact thought process when setting up Travis, but we likely did the minimum required to get it working.  Turning on extra features, like caching, would have been an extra improvement that we didn't immediately see a benefit from.

PR #1558, opened 2021-04-30
Sanitize keys before confirmation modal.

bamnet:
Without sanitization, these are susceptible to an varying degrees of XSS attack depending on the input.  When the values are displayed in the twitter-bootstrap-confirmation modal (I don't fully understand how it works) the JS is rendered and executed. The `user.name` fields are the most concerning for instances that allow public registration, but we sanitize everywhere this pattern is used to be safe.

Kudos to Ben Shaw (@sudonoodle) for finding this vulnerability!

### DIFF
--- a/app/views/concerto_plugins/show.html.erb
+++ b/app/views/concerto_plugins/show.html.erb
@@ -10,7 +10,7 @@
           <%= link_to t(:edit_model, model: ConcertoPlugin.model_name.human), edit_concerto_plugin_path(@concerto_plugin), class: "btn" %>
         <% end %>
         <% if can? :delete, @concerto_plugin %>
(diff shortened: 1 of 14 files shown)


bamnet:
@zr2d2 mind taking a look and making sure I've not messed too much up?

zr2d2 (commented):
These changes look good, but should we also be sanitizing user input before we put it into the database? I feel like not storing it in the database would be safer

bamnet:
While it would be safest, I considered that wouldn't resolve the problem if anyone had already put some XSS-triggering content in place and was just waiting for the right user, like an admin, to trigger it.

Names may also involve characters that the sanitize() function messes up, maybe bobby; drop tables has a sister <script>

zr2d2:
I suppose that's true, it does offer more protection if upgrading an existing deploy

PR #1563, opened 2021-10-23
use ruby 2.6

zr2d2:
@bamnet I checked, the fronted and the screen look fine. Is there a reason we didn't support 2.6 before?


bamnet:
I can't think of any reason why. Approving.

PR #1575, opened 2022-07-31
tidy up dockerfile, upgrade to ruby 2.6

zr2d2:
upgrades to ruby 2.6, inprove dockerfile

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,7 @@ services:
       - db:/var/lib/mysql
 
   concerto:
-    image: concerto
+    image: concertong
     restart: always
     ports: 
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -14,7 +14,7 @@ services:
       - db:/var/lib/mysql
 
   concerto:
-    image: concerto
+    image: concertong
```
jo-gross, line 17: Why have you changed this? If you like to change the name, you should update the README.md also.

zr2d2: good catch @jo-gross, that's not even a public image. I've reverted it in pr #1576

PR #1599, opened 2025-08-15
Arch 2025 Contributions

RandomGibus:
Restored refreshing, updated documentation

### DIFF
--- a/doc/Ruby_Installation_And_Maintanence_Guide.md
+++ b/doc/Ruby_Installation_And_Maintanence_Guide.md
@@ -12,7 +12,9 @@ Installing new ruby versions
     rvm pkg install openssl
 4. Install desired version\
     rvm install <ruby version here> --with openssl-dir=$rvm_path/usr
+
     if this fails, try running:\
-    rvm get head.
+    rvm get head
 
 Note: currently, Concerto runs on ruby 2.6.0
+


app/models/dynamic_content.rb
```diff
@@ -229,9 +229,16 @@ def destroy_children!
   # Find all the DynamicContent classes, find all the active content they have,
   # and then #{refresh} them.  Primarily invoked by our rake task.
   def self.refresh
-    dynamic_types = Concerto::Application.config.content_types.select do |t|
-      t.ancestors.include?(DynamicContent)
+    Rails.logger.debug("Refresh function called - !")
```
bamnet, line 232: Is it safe to remove these debug lines, here and below on 241?

doc/Calendar_Usage_Guide.md
```diff
@@ -0,0 +1,10 @@
+Using the Concerto calendar\
```
bamnet, line 1: Let's make this first line of markdown a header.

doc/Calendar_Usage_Guide.md
```diff
@@ -0,0 +1,10 @@
+Using the Concerto calendar\
+In order to Display a calendar on Concerto, you need to export it into a .ics format\
+Currently, there are very few ways to get a .ics link. Here's one way that I've found that works
+- Create a calendar in outlook
+- Share it with a given email(make sure to allow the email to view the full calendar, not just availability)
+- check response email, there will be a link underneath the button to connect the two emails.
+<img width="1567" height="688" alt="{D923B4E8-67D0-4F93-8430-D44701C6C3D6}" src="https://github.com/user-attachments/assets/a194b707-152d-4afd-85de-013535549c3b" />
```
bamnet, line 7: Can we add a more meaningful alt tag? I'm not sure what this UUID is.

doc/Clockwork installation.md
```diff
@@ -0,0 +1,5 @@
+Find concerto-clock.service (currently in concerto/script)\
```
bamnet, line 1: Like the other file, I suggest adding a `# header` or title to each markdown file before jumping straight into the content.

doc/Clockwork installation.md
```diff
@@ -0,0 +1,5 @@
...
+put file under /usr/lib/systemd/system (CentOS) or /lib/systemd/system (Ubuntu)\
+Run:\
+systemctl enable concerto-clock\
```
bamnet, line 4: Wrap commands in backticks so they get styled properly in markdown. I think there's an opportunity to do this throughout a handful of files, like the rebooting instructions etc

doc/Clockwork installation.md
bamnet, line 1: I've noticed a lot of markdown lines end with a trailing backslash.  I think we should either continue the line in place OR use two blank lines to indicate a new paragraph.  It's going to be kind of annoying to manually maintain these backlashes at the end of every sententence.

So the markdown should look either:

```md
Like this with the first sentence.
And the second sentence follows here
```

OR

```md
The first sentence here.

The second sentence down here.
```

script/concerto-clock.service
bamnet, line 1: Honest question (I don't know the answer!) - Do changes to either of these script .service files impact how other people might be deploying Concerto 2?

Is there anything we need to change in [Dockerfile](https://github.com/concerto/concerto/blob/master/Dockerfile), [docker-compose.yml](https://github.com/concerto/concerto/blob/master/docker-compose.yml), [concerto-init.d](https://github.com/concerto/concerto/blob/master/concerto-init.d) or alike?

bamnet (changes requested):
Thanks for kicking off this PR!

I've added a mix of comments mainly around some areas the Markdown could be improved, and also where I'd like to understand a bit more on the deployment impact.

app/models/dynamic_content.rb
RandomGibus: yes, removing those should be fine

PR #1672, opened 2026-03-08
Add inline re-moderation UI on content show page

### LINKED ISSUE #1667: Update content moderation status after initial moderation
After content has been moderated (e.g. approved or rejected), there's no way for a moderator to go back and re-moderate it.

This is particularly important if a feed moderator approved something and later realizes they should have rejected or otherwise needs to reject it.

Maybe we should consider some inline moderation UI on the content#show page for moderators?

bamnet:
## Summary

- Adds a "Change" button next to each submission's status badge on the content show page, visible only to moderators
- Clicking "Change" expands an inline form (hidden by default, scoped per submission) with Approve and Reject actions — uses the existing `toggle` Stimulus controller
- The reject form includes an optional reason textarea, pre-filled when updating an existing rejection
- `moderate` action now uses `redirect_back` so it returns to the content page when triggered from there, falling back to the moderation queue otherwise
- Status badge is now always shown (not just for non-approved) to give context for the "Change" button

## Test plan

- [x] New controller tests: re-reject approved, re-approve rejected, non-member blocked from re-moderation, redirect back to referer, fallback redirect
- [x] Updated existing test asserting "Approved" badge is now shown
- [x] Full test suite passes (870 tests, 0 failures)

Closes #1667

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -21,7 +21,7 @@ class SubmissionsController < ApplicationController
 
     unless %w[approved rejected].include?(status)
       return respond_to do |format|
-        format.html { redirect_back_or_to submissions_path, alert: "Invalid moderation action." }
+        format.html { redirect_back_or_to submissions_path, allow_other_host: false, alert: "Invalid moderation action." }
         format.json { render json: { error: "Invalid moderation action" }, status: :bad_request }
       end
     end
@@ -29,7 +29,7 @@ class SubmissionsController < ApplicationController
     @submission.moderate!(status: status, moderator: current_user, reason: reason)
 
     respond_to do |format|
-      format.html { redirect_back_or_to submissions_path, notice: "Submission was #{status}." }
+      format.html { redirect_back_or_to submissions_path, allow_other_host: false, notice: "Submission was #{status}." }
       format.json { render json: @submission, status: :ok }
     end
   end


app/controllers/submissions_controller.rb
```diff
@@ -21,15 +21,15 @@ def moderate
 
     unless %w[approved rejected].include?(status)
       return respond_to do |format|
-        format.html { redirect_to submissions_path, alert: "Invalid moderation action." }
+        format.html { redirect_back fallback_location: submissions_path, alert: "Invalid moderation action." }
```
bamnet: redirect_back is soft deprecated and just a wrapper around [redirect_back_or_to](https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_back_or_to). Let's swap to that.

ref:
```ruby
def redirect_back(fallback_location:, allow_other_host: _allow_other_host, **args)
  redirect_back_or_to fallback_location, allow_other_host: allow_other_host, **args
end
```

app/controllers/submissions_controller.rb
```diff
@@ -21,15 +21,15 @@ def moderate
...
         format.json { render json: { error: "Invalid moderation action" }, status: :bad_request }
       end
     end
 
     @submission.moderate!(status: status, moderator: current_user, reason: reason)
 
     respond_to do |format|
-      format.html { redirect_to submissions_path, notice: "Submission was #{status}." }
+      format.html { redirect_back fallback_location: submissions_path, notice: "Submission was #{status}." }
```
bamnet: (same here)

app/views/submissions/_moderation_actions.html.erb
```diff
@@ -0,0 +1,36 @@
+<%# Inline re-moderation form for a single submission.
+    Expects: submission
+    Outer div is scoped to this submission's toggle controller (show/hide reject form). %>
```
bamnet: We don't typically add comments like this.  While I appreciate the enthusiasm, I don't think it's needed here.

app/views/submissions/_moderation_actions.html.erb
bamnet, line 1: Is there an opportunity to re-use this partial and DRY up app/view/submissions/_card.html.erb which also offers an inline moderation experience (albeit not within a collapsible div)?

app/views/submissions/_moderation_actions.html.erb
```diff
@@ -0,0 +1,36 @@
...
+        <%= form.hidden_field :moderation_status, value: "approved" %>
+        <%= form.submit "Approve", class: "btn-primary w-full cursor-pointer text-sm" %>
+      <% end %>
+    <% end %>
+
+    <button type="button"
+            data-action="toggle#toggle"
+            class="btn-danger flex-1 text-sm">
+      <%= submission.rejected? ? "Update Rejection" : "Reject" %>
+    </button>
```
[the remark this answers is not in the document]
bamnet: I didn't notice this in my testing, but maybe I didn't have a fancy enough example.  Do we need to worry about this anywhere else the Toggle controller is used?

PR #1686, opened 2026-03-13
Add files via upload

CIAQ0404:
Created a new python file to take given JSON file containing weather information and turning that into a HTML file, so weather can be properly displayed.

### DIFF
--- a/Weather_HTML.py
+++ b/Weather_HTML.py
@@ -1,0 +1,207 @@
+import json
+import time
+import os
+from datetime import datetime
+
+# =========================
+# CONFIG
+# =========================
+JSON_FILE = "weather.json"      # local JSON file being updated by another service
+OUTPUT_HTML = "weather.html"    # HTML file Concerto will display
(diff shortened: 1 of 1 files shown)


bamnet:
Hi @CIAQ0404!

Modern weather support would be a great addition to Concerto!  It would be interesting to prototype this using our new Remote Feed capability. Remote Feeds are a way for Concerto to poll a JSON endpoint which returns Content items - like the HTML page you're generating here, and make that content available to screens.

This approach avoids the need to distribute a specific weather provider in Concerto itself (there are pretty wide regional variations, and some licensing complexity), while making it easy for Concerto to  constantly pull in the latest and greatest content.  It currently powers our "Quota of the day" service, which you can see @ https://demo.concerto-signage.org and my https://github.com/bamnet/qod repo.

https://www.concerto-signage.org/docs/remote-feeds/ has some documentation which might help you get started.  Feel free to open an issue or discussion to chat more about this!

PR #1762, opened 2026-04-24
Respect video aspect ratio on the player

### LINKED ISSUE #1761: Respect aspect ratio of video content
When video content is playing, like a YouTube video, the iframe player typically stretches to fill the entire position.  The video is often not the same aspect ratio as the position, resulting in some black margins around the video.  I don't think this looks particularly good.

Instead, it would be great if the player had some way of staying constrained to the video's aspect ratio or had  a transparent background so the margin wasn't as eye catching.

For example, observe all the extra black around the YouTube video being played here: 

Image

We might need different implementations for each video provider?

bamnet:
Fixes #1761.

## Summary

Video iframes no longer stretch to fill the field. Each video component now wraps the iframe in a centered container and constrains it to an aspect ratio, so the surrounding space shows the template background instead of the embed player's black letterbox.

Ratios are computed server-side (`Video#effective_aspect_ratio`) with these defaults:
- **YouTube:** `16/9`, or `9/16` when the URL contains `/shorts/`
- **Vimeo:** `16/9`, then dynamically corrected on the frontend via `player.getVideoWidth()` / `getVideoHeight()` when the user hasn't set an override
- **TikTok:** `9/16`

YouTube's IFrame API has no equivalent to Vimeo's dimension getters (confirmed against the current API reference), and YouTube oEmbed always returns `200x113` regardless of the video's true orientation — so there's no reliable dynamic signal. To handle the remaining case (e.g. a vertical Short shared as a `watch?v=` link), a new "Aspect Ratio" selector on the video form lets the author override the default per video.

## Design notes

- New `Video::ASPECT_RATIOS` whitelist: `auto`, `16:9`, `9:16`, `4:3`, `1:1` (stored in the existing `config` JSON via `store_accessor`, no migration needed).
- `Video#as_json` exposes `aspect_ratio` (CSS-ready, e.g. `"9/16"`) and `aspect_ratio_auto` (bool) so the Vimeo component knows when it's safe to override from the Player API.
- Provider default/Shorts-URL logic lives in the model so the frontend stays thin.

## Test plan
- [x] `bin/rails test` (917 runs, 0 failures)
- [x] `yarn run vitest run` (13 files, 119 passed / 1 skipped)
- [x] `bundle exec rubocop` (no offenses)
- [x] `yarn run eslint "{app,test}/frontend/**/*.{js,vue}"` (clean)
- [x] Model tests cover per-provider defaults, Shorts URL detection, user override, `auto` fallback, validation.
- [x] Controller tests cover `aspect_ratio` permit/reject.
- [x] Frontend tests verify iframe `aspect-ratio` style per provider, and Vimeo dynamic correction only when `aspect_ratio_auto` is true.
- [ ] Manual browser smoke-test with a regular 16:9 YouTube, a `/shorts/` URL, a vertical Vimeo, a TikTok, and a manually-overridden video — confirm letterboxing against template background rather than black bars.

### DIFF
--- a/app/models/video.rb
+++ b/app/models/video.rb
@@ -4,10 +4,18 @@ require "cgi"
 class Video < Content
   store_accessor :config, :url, :aspect_ratio
 
-  # Permitted values for the user-facing aspect ratio override.
-  # nil / "auto" means "use the provider default" (with /shorts/ URL detection
-  # for YouTube and dynamic correction for Vimeo on the frontend).
-  ASPECT_RATIOS = [ "auto", "16:9", "9:16", "4:3", "1:1" ].freeze
+  # Permitted values for the user-facing aspect ratio override, with the
+  # labels shown in the form. nil / "auto" means "use the provider default"
+  # (with /shorts/ URL detection for YouTube and dynamic correction for Vimeo
+  # on the frontend).
+  ASPECT_RATIO_OPTIONS = {
+    "auto" => "Auto (detect from source)",
+    "16:9" => "16:9 (widescreen)",
+    "9:16" => "9:16 (vertical)",
+    "4:3" => "4:3 (standard)",
+    "1:1" => "1:1 (square)"
+  }.freeze
+  ASPECT_RATIOS = ASPECT_RATIO_OPTIONS.keys.freeze
 
   validates :aspect_ratio, inclusion: { in: ASPECT_RATIOS }, allow_blank: true
 
@@ -25,10 +33,10 @@ class Video < Content
   # user has not set an explicit override.
   def effective_aspect_ratio
     ratio = aspect_ratio
-    return css_ratio(ratio) if ratio.present? && ratio != "auto"
+    return ratio.tr(":", "/") if ratio.present? && ratio != "auto"
 
     case video_source
-    when "youtube" then youtube_default_ratio
+    when "youtube" then url.to_s.include?("/shorts/") ? "9/16" : "16/9"
     when "tiktok" then "9/16"
     else "16/9"
     end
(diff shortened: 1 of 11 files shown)


app/models/video.rb
```diff
@@ -78,6 +101,14 @@ def thumbnail_url
 
   private
 
+  def css_ratio(ratio)
```
bamnet: This function feels like something we should just move inline since it's only used once AFAICT.

app/views/videos/_form.html.erb
```diff
@@ -39,6 +39,24 @@
...
+                [ "Auto (detect from source)", "auto" ],
+                [ "16:9 (widescreen)", "16:9" ],
+                [ "9:16 (vertical)", "9:16" ],
+                [ "4:3 (standard)", "4:3" ],
+                [ "1:1 (square)", "1:1" ]
+              ],
+              { selected: video.aspect_ratio.presence || "auto" },
+              class: "block w-full px-3 py-2 border border-gray-300 rounded-lg shadow-sm focus:outline-none focus:ring-2 focus:ring-blue-500 focus:border-blue-500 text-sm" %>
+          <p class="text-xs text-gray-500 mt-1">
+            Override when auto-detection is wrong — for example, YouTube Shorts shared via a <code>watch?v=</code> link.
```
bamnet: instead of "shared ...", let's use "linked via a ... URL"

app/views/videos/_form.html.erb
```diff
@@ -39,6 +39,24 @@
           </p>
         </div>
 
+        <!-- Aspect Ratio -->
+        <div>
+          <%= form.label :aspect_ratio, "Aspect Ratio", class: "block text-sm font-medium text-gray-700 mb-2" %>
+          <%= form.select :aspect_ratio,
```
bamnet, line 45: Is there some way we can keep this list in sync with ASPECT_RATIOS in the model? Ideally we don't define it twice.

PR #1769, opened 2026-04-25
Support PDF uploads for Graphic content

### LINKED ISSUE #1764: Support PDF uploads
It would be great if users could upload flyers which were PDFs.

Behind the scenes, I think we'd want to convert the first page of the PDF to an image and model it as a piece of Graphic content. This was functionality we had in previous versions of Concerto, and it would be great to figure out how to carry it forward.

I vaguely remember that the PDF support in Concerto 1 had a painful set of dependencies, but given our Docker-based default deployment strategy this may be easier to address.

(initially reported by @plgone in #1763 )

bamnet:
Closes #1764

## Summary

- Users can upload a PDF as a Graphic; the first page is converted to a PNG asynchronously via `ConvertPdfToImageJob` using `ImageProcessing::Vips` (requires libvips + poppler)
- The admin show page displays a processing spinner while conversion is pending, or an inline error message if it fails (stored in `config` as `conversion_error`)
- `Graphic#should_render_in?` returns `false` while the image is still a PDF, so the frontend player never receives an unconverted graphic
- The upload form now accepts `image/*,application/pdf`

## Dependencies

libvips must be compiled with poppler support. The Dockerfile already has `libvips`; non-Docker Debian/Ubuntu users need:
```
sudo apt install libvips-dev libpoppler-glib-dev
```

## Test plan

- [ ] Upload a PDF via the Graphic upload form and confirm the processing spinner appears immediately
- [ ] Refresh after the background job runs and confirm the converted image is displayed
- [ ] Verify the graphic does not appear on the player while still a PDF
- [ ] Confirm a regular image upload still works normally
- [ ] `bin/rails test` passes


app/views/graphics/_form.html.erb
```diff
@@ -33,8 +33,8 @@
           <%= form.label :image, "Image File", class: "block text-sm font-medium text-gray-700 mb-2" %>
           <%= form.file_field :image,
               class: "block w-full text-sm text-gray-500 file:mr-4 file:py-2 file:px-4 file:rounded-lg file:border-0 file:text-sm file:font-medium file:bg-blue-50 file:text-blue-700 hover:file:bg-blue-100 border border-gray-300 rounded-lg cursor-pointer focus:outline-none focus:ring-2 focus:ring-blue-500 focus:border-blue-500",
-              accept: "image/*" %>
-          <p class="text-xs text-gray-500 mt-1">Supported formats: JPG, PNG</p>
+              accept: "image/*,application/pdf" %>
+          <p class="text-xs text-gray-500 mt-1">Supported formats: JPG, PNG, PDF (first page will be used)</p>
```
bamnet: I don't think we need the detail about first page being used. Just PDF is OK.

app/views/graphics/_grid.html.erb
```diff
@@ -1 +1,5 @@
-<%= image_tag content.image.variant(:grid), class: "object-contain" %>
\ No newline at end of file
+<% if content.image.attached? && content.image.content_type != "application/pdf" %>
```
bamnet: I see the `content_type != "application/pdf"` logic repeated a few times.  Is there an opportunity to add a method to Graphic which indicates whether it's still processing or not?  That way, if we add another content_type that gets processed, beyond PDF, we don't have to update this views.

PR #1770, opened 2026-04-25
Validate Graphic image content type, fix 500 on unsupported files

### LINKED ISSUE #1768: Return error message for unsupported Graphic content filetype
If users attempt to upload an unsupported file as a Graphic, Concerto returns a 500 error.  This is not very user friendly. We should return a clear error to the user explaining that the file format is unsupported.

bamnet:
Closes #1768

## Summary

- Unsupported file uploads are now rejected at the model layer with a clear validation error ("Image is not a supported format") — the file is never stored
- The allowlist is `ActiveStorage.variable_content_types + ["application/pdf"]`, so it stays current with whatever Rails/libvips supports without code changes
- Defense-in-depth: `_graphic.html.erb` now shows an "Unsupported file format" placeholder instead of 500ing on records already in the database; `_grid.html.erb` and `should_render_in?` use `image.variable?` for the same reason

## Test plan

- [ ] Upload an unsupported file (e.g. `.exe`, `.zip`) — expect a form validation error, no file stored
- [ ] Upload a supported image (JPG, PNG, WebP) — expect normal behaviour
- [ ] Upload a PDF — expect conversion spinner as before
- [ ] `bin/rails test` passes

### DIFF
--- a/app/models/graphic.rb
+++ b/app/models/graphic.rb
@@ -5,6 +5,8 @@ class Graphic < Content
 
   store_accessor :config, :conversion_error
 
+  SUPPORTED_CONTENT_TYPES = (ActiveStorage.variable_content_types + [ "application/pdf" ]).freeze
+
   # URL Helpers are needed so we can generate a URL to the image in the JSON.
   include Rails.application.routes.url_helpers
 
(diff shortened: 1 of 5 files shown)


app/models/graphic.rb
```diff
@@ -67,4 +71,9 @@ def convert_pdf_to_image_if_needed
     update_column(:config, (config || {}).except("conversion_error")) if conversion_error.present?
     ConvertPdfToImageJob.perform_later(self)
   end
+
+  def image_content_type_supported
+    return if SUPPORTED_CONTENT_TYPES.include?(image.content_type)
+    errors.add(:image, "is not a supported format")
```
bamnet: Can we reference the content type in the error message, so users know what we checked?

Issue #1689, opened 2026-03-18
Fix template position borders


bamnet:
Many Concerto 1.x templates include CSS styling like a border.

In our current implementation, this is set on the position element which is typically larger than the content being displayed.  As a result, the border is floating off around the outline of the position instead of bordering the content itself.  This looks bad.

Three options:
1. Apply the style to a container that's tightly sized around the content itself
2. Push some styles, like borders, to the content being rendered, instead of the larger container. Let content figure out if they should apply the styles or not.  For non-box content, like RichText, a border might not make sense.
3. Just remove these borders from the current templates. This would allow us to support a future where a position really should have a border.

Many of the current templates were built for a world where fields showed one type of content, like graphics in the main field, where a border made sense.  Now that we support content in a wider variety of fields, I'm not sure how many of the borders make sense.

PR #1780, opened 2026-04-27
Scale graphics to fill their container

### LINKED ISSUE #1779: Smaller graphics don't scale to fit frontend
If a graphic is smaller than the position it's being rendered in, it just sits in the center of the position instead of trying to stretch and fill it while preserving the aspect ratio.

For example I uploaded a flyer which, after PDF conversion, has a height of 1275.  If the position is larger than that - like it might be showing BlueSwoosh on a 4k screen, the flyer has lots of extra space around all sides.

bamnet:
## Summary
- `ConcertoGraphic` used `max-width: 100%; max-height: 100%`, which clamps oversized images down but never grows undersized ones. A 1275px-tall PDF flyer therefore rendered at native pixel size on a 4K screen, leaving wide letterbox/pillarbox margins.
- Switching to `width: 100%; height: 100%` (paired with the existing `object-fit: contain`) makes the image scale up or down to fill the container while preserving aspect ratio. Mismatched aspect ratios still letterbox rather than crop.
- Added a test asserting the computed style of `.graphic`.

Fixes #1779.

## Trade-offs

- Low-resolution graphics on big screens will get upscaled and look softer. That's the explicit ask in the issue — preserving aspect ratio without crop means accepting some upscaling.
- A `boxStyle` border (the per-template border on the main field) now outlines the full field rectangle rather than hugging the image. This matches how the video components already behave (`ConcertoYoutubeVideo` et al. apply `boxStyle` to a `width: 100%; height: 100%` wrapper), so we're not introducing a new inconsistency, just propagating the existing one to graphics. The proper fix is to solve sizing & borders once at the field/composable level for any content type with an aspect ratio — tracked in #1781.

We initially tried two patches that hugged the image (pure-CSS aspect-ratio and a JS `fitImage()` + ResizeObserver). The CSS one stretched images under `max-height` clamping; the JS one worked but parked complexity in a single component while leaving videos unchanged. Both are documented in #1781 for reference.

## Test plan
- [x] `yarn run vitest` (123 tests, 1 skipped)
- [x] `yarn run eslint` clean
- [x] Manually verified small graphics now fill large positions on the dev environment

### DIFF
--- a/app/frontend/components/ConcertoGraphic.vue
+++ b/app/frontend/components/ConcertoGraphic.vue
@@ -1,16 +1,61 @@
 <script setup>
+import { ref, onMounted, onBeforeUnmount } from 'vue'
+
 defineProps({
   content: {type: Object, required: true},
   boxStyle: {type: String, required: false, default: ''},
 });
+
+const containerRef = ref(null)
+const imgRef = ref(null)
+let resizeObserver = null
+
+// Constrain the binding axis to 100% and let the image's intrinsic
+// aspect ratio drive the other dimension. The img element box ends up
+// equal to the rendered content rectangle, so a boxStyle border hugs
+// the image at any container size — including when the image has been
+// scaled up to fill a position larger than its native pixel dimensions.
+function fitImage() {
+  const container = containerRef.value
+  const img = imgRef.value
+  if (!container || !img || !img.naturalWidth || !img.naturalHeight) return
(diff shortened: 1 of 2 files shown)


bamnet:
This is all getting quite complicated. Maybe we should stick with the earlier fix dropping the "max-" prefix from CSS. This will break borders, but I'm pretty sure borders aren't working for videos anyways. We probably need to re-open #1689.

PR #1786, opened 2026-04-29
Render videos in portrait positions

### LINKED ISSUE #1784: Videos don't like to render in portrait screens
Like #1778, videos seem equally unhappy rendering in the 'main' field on an average portrait screen.  I think we need to adjust our logic here, especially since a 9:16 video would look great on a tall position on a portrait screen.

bamnet:
## Summary
- `Video#should_render_in?` was hardcoded to `(0.25 < ar <= 1)`, which ignored the video's own aspect ratio. It rejected anything wider than square in template-normalized coords and didn't differentiate between a 16:9 widescreen and a 9:16 short.
- Mirror the content-aware pattern from `Graphic#should_render_in?`: compare the position's aspect ratio to the video's `effective_aspect_ratio` within a configurable tolerance (4×).
- A 9:16 short now renders happily in tall sidebars; landscape videos still get rejected from extreme tickers and very narrow portrait positions; the player continues to letterbox cross-orientation cases.

## Tolerance choice (4×)
We considered three multipliers; recording for future tuning:

| Multiplier | 16:9 in `two_graphic` (ar 0.74) | 9:16 in 16:9 main (ar 1.78) | Tickers (ar 7.5) |
|---|---|---|---|
| 2× | rejected (regresses existing test) | rejected | rejected |
| 3× | fits | rejected (just barely) | rejected |
| **4×** | **fits** | **fits** | **rejected (just barely)** |

4× wins for now because it allows full cross-orientation placement (the issue specifically calls out wanting 9:16 shorts to play in landscape mains and vice versa) while still excluding tickers. The downside is little headroom — anything wider than a ticker would slip through. If we pick up wide banner positions later we may want to tighten back to 3× and add an explicit position-shape check.

Fixes #1784.

## Test plan
- [x] `bin/rails test test/models/video_test.rb`
- [x] `bin/rails test` (full suite)
- [x] `bundle exec rubocop app/models/video.rb test/models/video_test.rb`
- [ ] Manually verify a 9:16 video plays in a portrait main and a 16:9 video plays in a portrait main with letterboxing.

### DIFF
--- a/app/models/video.rb
+++ b/app/models/video.rb
@@ -49,7 +49,8 @@ class Video < Content
   # letterboxes to fit. A tighter 2x-3x bound is more shape-aware but rejects
   # useful cross-orientation cases.
   def should_render_in?(position)
-    ratio = effective_aspect_ratio_value
+    width, height = effective_aspect_ratio.split("/").map(&:to_f)
+    ratio = width / height
     (position.aspect_ratio / 4.0) <= ratio && ratio <= (position.aspect_ratio * 4.0)
   end
 
@@ -111,11 +112,6 @@ class Video < Content
   end
 
   private
-
-  def effective_aspect_ratio_value
-    width, height = effective_aspect_ratio.split("/").map(&:to_f)
-    width / height
-  end
 
   # Memoized helper to fetch Vimeo oEmbed data once per request
   def vimeo_oembed_data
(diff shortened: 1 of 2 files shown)


app/models/video.rb
```diff
@@ -109,6 +112,11 @@ def thumbnail_url
 
   private
 
+  def effective_aspect_ratio_value
```
bamnet: Since this function is only used once, I suggest moving it inline.

PR #1787, opened 2026-04-29
Add cross-model search via SQLite FTS5

### LINKED ISSUE #1785: Add content search
As a Concerto installation grows, it would be helpful to have the ability to search for content.  Currently, users are limited to scrolling around on content#index which is great when there are only a dozen pieces of content but not particularly practical for installations with hundreds of pieces of content.

Two questions to consider:
1) Should we reuse the top level placeholder search box which currently says "Search content..." or should we add a dedicated Search box specifically for content to content#index?
2) We want to add the ability to search other models in the future - like feeds.  How does that impact our implementation plan?

I read that Rails + sqlite3 recently gained some basic full text search capabilities using virtual tables, that might be an interesting implementation approach.  It would be cool if we could index inside content, like graphics, but I have a feeling that is biting off more than we should chew for this feature.

bamnet:
## Summary

- Polymorphic SQLite FTS5 corpus indexes Content (RichText, Graphic, Video, Clock) and Feed (RssFeed, RemoteFeed). Each STI subclass declares its own searchable text — RichText body has HTML stripped, Video indexes its URL, Clock its format, Graphic its filename.
- New `/search` route for cross-model results from the header search box; `?q=` on `/contents` and `/feeds` scopes search to one model and composes with the existing `?scope=` toggle on `contents#index`.
- Unapproved Content is excluded from the index entirely (so result counts aren't eaten by invisible rows). A `Submission` `after_commit` hook reindexes the parent Content when moderation flips. Pundit `policy_scope` is still applied post-fetch as defense in depth.
- User input never reaches the FTS5 parser: `Search.build_match` tokenizes on `\W+`, quotes each token, and appends a prefix wildcard.
- `bin/rails search:reindex` rebuilds the corpus; `search:verify` prints any drift between source rows and corpus rows per model.

Closes #1785.

## Design notes

- Single shared FTS5 virtual table (`search_corpus`) with `searchable_type`/`searchable_id` UNINDEXED columns. STI is collapsed to base class (`Content`/`Feed`) so a row hit resolves to the right subclass via Rails' STI.
- Migration uses Rails 7.1+ `create_virtual_table`, which round-trips into `db/schema.rb` — no `structure.sql` switch needed and no test_helper shim to recreate the table.
- `Searchable` is a thin concern (`after_commit` reindex + `searchable_data` interface). Each model overrides `searchable_data` and may override `searchable?` to gate inclusion.
- Search has two entry points: `Search.call(q, user:)` for global ranked results (returns flat Array, policy_scoped) and `Search.matching_ids(q, klass)` for the per-page case where the caller composes its own AR scopes.
- Architectural decisions, alternatives considered, and verification steps are captured in the planning doc this branch was scaffolded from.

## Out of scope (intentional, called out for future work)

- Screen searchability — `Searchable` makes adding it a 3-line change; deferred to keep this PR focused.
- Indexing inside Graphic image content (OCR) and oEmbed video metadata.
- Late-attached Graphic image filename: a Graphic created without an image and attached afterward indexes by name only (the ActiveStorage attachment commit is separate). Documented; small enough to fix in a follow-up if needed.
- Live search / autocomplete.
- bm25 weight tuning beyond defaults (`name=10.0`, `body=1.0`).

## Test plan

- [x] `bin/rails test` — full suite green (969 runs, 0 failures)
- [x] `yarn run vitest` — frontend suite green (122 passed)
- [x] `bin/rubocop` clean
- [x] `yarn run eslint "{app,test}/frontend/**/*.{js,vue}"` clean
- [x] `bin/rails search:reindex` populates corpus from seed data
- [x] `bin/rails search:verify` shows zero drift
- [ ] Browser smoke: header search submits to `/search`, returns mixed Content + Feed results
- [ ] Browser smoke: `/contents?q=…` narrows results, `?scope=upcoming&q=…` correctly intersects, "Clear" link removes the query
- [ ] Browser smoke: `/feeds?q=…` narrows feed list
- [ ] Browser smoke: approve a pending Submission via the moderation UI → its Content appears in search; reject it → it disappears
- [ ] Browser smoke: adversarial queries (`"`, `*`, `OR x`, `NEAR(foo,5)`, very long input, empty) don't error or leak FTS5 syntax behavior


app/models/content.rb
```diff
@@ -12,6 +14,17 @@ def self.policy_class
...
 
+    # Searchable: only index Content with at least one approved submission so
+    # unapproved drafts don't consume search-result slots that policy_scope
+    # would later filter out.
+    def searchable?
+      submissions.where(moderation_status: :approved).exists?
+    end
+
+    def searchable_data
+      { name: name, body: text }
```
bamnet: > body: text

I don't think content has a meaningful text value here. This is something we should rely on sub-classes to define.

app/models/remote_feed.rb
```diff
@@ -33,6 +33,10 @@ def refresh_interval
       super.to_i if super
     end
 
+    def searchable_data
+      { name: name, body: [ description, url ].compact_blank.join(" ") }
```
bamnet: url: Can we strip off the query parameters in case those contain secret keys?

For example, I think we should index https://foo.com/feed.json, not https://foo.com/feed.json?key=ABC123

app/models/rss_feed.rb
```diff
@@ -30,6 +30,10 @@ def render_mode
       (ticker? ? RichText.render_as[:plaintext] : RichText.render_as[:html])
     end
 
+    def searchable_data
+      { name: name, body: [ description, url ].compact_blank.join(" ") }
```
bamnet: Like my comment on remote_feed, can we strip query parameters before indexing URLs?

For example, I think we should index https://foo.com/feed.json, not https://foo.com/feed.json?key=ABC123

app/models/video.rb
```diff
@@ -97,6 +97,10 @@ def video_source
     end
   end
 
+  def searchable_data
+    { name: name, body: url.to_s }
```
bamnet, line 101: Note - Unlike RemoteFeed and RSS Feed, I don't think this is likely to contain any secrets, so we should index the full thing.

app/services/search.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  # Scoped, composable. Returns Array<Integer> of ids matching `klass` (keyed
+  # on its base class in the corpus). Caller applies its own policy_scope and
+  # any additional ActiveRecord scopes — used by ContentsController and
+  # FeedsController so search composes with the existing `?scope=` toggle.
+  def self.matching_ids(query, klass, limit: 200)
+    match = build_match(query)
+    return [] if match.blank?
+
+    sql = ActiveRecord::Base.sanitize_sql_array([
```
bamnet: I see we're hardcoding some SQL here. That's usually discouraged in Rails apps.  Did we consider any ActiveRecord based implementations here?

app/views/search/index.html.erb
```diff
@@ -0,0 +1,60 @@
...
+<% end %>
+
+<% if @query.present? && @results.any? %>
+  <div class="space-y-3">
+    <% @results.each do |record| %>
+      <%= link_to polymorphic_path(record), class: "block card hover:shadow-md transition-shadow duration-200" do %>
+        <div class="card-body">
+          <div class="flex items-start justify-between gap-4">
+            <div class="flex-1 min-w-0">
+              <div class="flex items-center gap-2 mb-1">
```
bamnet, line 30: Can we make the record type a bit more obvious? Maybe with some color or iconography?  Scanning the list, it's quite difficult to tell what is a piece of content vs what is a feed. The string names "RichText" and "RSSFeed" are so close.

app/views/search/index.html.erb
```diff
@@ -0,0 +1,60 @@
...
+                  <%= record.class.name %>
+                </span>
+              </div>
+              <h3 class="text-h3 text-neutral-900 truncate">
+                <%= record.name.presence || "(untitled)" %>
+              </h3>
+              <% snippet = record.searchable_data[:body].to_s.truncate(160) %>
+              <% if snippet.present? %>
+                <p class="text-body text-neutral-700 mt-1 truncate">
+                  <%= snippet %>
```
bamnet, line 39: Is there any way for us to highlight the text matching in the snippet? It's OK if not.

bamnet:
Pushed ca0ddef addressing the review comments:

- **`Content#searchable_data`**: dropped from the base class. Each STI subclass already defines its own — Content base no longer pretends to know about `body`.
- **RSS / Remote feed URLs**: query string and fragment stripped before indexing via a private `Feed#indexable_url` helper. So `https://foo.com/feed.json?key=ABC123` indexes as `https://foo.com/feed.json`. Video URLs left intact per your note.
- **Raw SQL → ActiveRecord**: introduced a `SearchRow < ApplicationRecord` model bound to the FTS5 virtual table. `Search::Corpus` now uses AR for delete/upsert (wrapped in a transaction, also addressing the gemini-code-assist comment) and counts; `Search` uses AR relations with raw fragments only for the FTS5-specific `MATCH` operator and `bm25()` ranking. Inserts use `insert_all!` to sidestep FTS5's magic columns (`search_corpus`, `rank`) and the absent `id` PK.
- **Type badges**: per-subclass icon and base-class color (blue for Content, amber for Feed) plus a readable label ("Text", "Graphic", "RSS Feed", …). At-a-glance distinction without relying on close-looking class names.
- **Match highlighting**: skipped per your "OK if not".

Tests: 969 Rails + 122 frontend, all passing. RuboCop and ESLint clean.

app/services/search.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+module Search
+  DEFAULT_LIMIT = 50
+  RANK_ORDER = Arel.sql("bm25(search_corpus, 10.0, 1.0)")
+
+  # Global ranked cross-model search. Returns a flat Array of records ordered
+  # by FTS5 rank, already filtered through Pundit policy_scope per registered
+  # type.
+  def self.call(query, user:, types: Corpus.registry, limit: DEFAULT_LIMIT)
```
bamnet, line 10: This `user:,` syntax looks a bit weird to me.  Is there a clearer way to express this?

bamnet:
### Code review

Found 2 issues:

1. `SearchController` is missing `after_action :verify_authorized` and `after_action :verify_policy_scoped`, and has no associated `SearchPolicy`. CLAUDE.md requires: "Every controller action must call `authorize @record` or `policy_scope(Model)`" and "Use `after_action :verify_authorized` and `after_action :verify_policy_scoped` in controllers." While `Search.call` internally calls `Pundit.policy_scope!`, this does not satisfy the controller-level Pundit contract the rest of the codebase enforces. The controller should either add a `SearchPolicy` with the appropriate hooks, or call `skip_authorization`/`skip_policy_scope` with a documented reason.

https://github.com/concerto/concerto/blob/ca0ddef1bef359ff771c185d8427b779f6d38810/app/controllers/search_controller.rb#L1-L6

2. Inline SVG search icons in the new/modified views use `stroke-width="2"` instead of the `1.5` required by CLAUDE.md ("uses line icons with 1.5px stroke width"). The empty-state icon in the same file correctly uses `1.5`. The `heroicon` helper at `app/helpers/application_helper.rb` already handles this correctly and could be used instead.

https://github.com/concerto/concerto/blob/ca0ddef1bef359ff771c185d8427b779f6d38810/app/views/search/index.html.erb#L17-L19

https://github.com/concerto/concerto/blob/ca0ddef1bef359ff771c185d8427b779f6d38810/app/views/contents/index.html.erb#L64-L66

🤖 Generated with [Claude Code](https://claude.ai/code), but edited by Brian

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1813, opened 2026-05-12
Surface uploader's content regardless of moderation state

### LINKED ISSUE #1798: Users cannot see their uploaded content
As a user, I should be able to see all of the content that I have uploaded, regardless of the moderation state or the number of feeds it's been submitted to.

Currently, this isn't possible. All the list views (content#index, feed#show) use breakdowns which focus on moderated content.  As a result, if your content is pending / rejected or not yet submitted to any feeds it's very hard to discover.

bamnet:
## Summary
- `ContentPolicy::Scope` now ORs in `where(user_id: user.id)` so signed-in users see their pending/rejected/unsubmitted content alongside the public approved set.
- `/contents` gains a **Mine** tab (and matching **My Content** sidebar shortcut) showing every upload by the current user, with `Pending` / `Rejected` / `Not in feeds` badges in the grid.
- `/feeds/:id` gains a **Mine** tab, visible to any signed-in user, that lists their submissions across all moderation statuses. Non-moderator owners see "Awaiting moderation by a feed manager" instead of the approve/reject form.

Fixes #1798.

## Test plan
- [x] `bin/rails test` — 990 runs, 0 failures (6 new policy-scope tests, 7 new contents-controller tests, 6 new feeds-controller tests; existing search test adjusted to use a non-owner viewer)
- [x] `bundle exec rubocop` clean
- [x] `yarn run vitest run` — 122 pass
- [x] `yarn run eslint` clean
- [x] Manual smoke (Playwright): signed in, sidebar shows **My Content**; `/contents?scope=mine` lists owned items including an unsubmitted draft with a `Not in feeds` badge; `/feeds/1?status=mine` renders the Mine tab with the empty-state CTA.

### DIFF
--- a/app/controllers/contents_controller.rb
+++ b/app/controllers/contents_controller.rb
@@ -10,23 +10,8 @@ class ContentsController < ApplicationController
     @scope = "active" if @scope == "mine" && !user_signed_in?
     @query = params[:q].to_s.strip
 
-    contents_scope = case @scope
-    when "active"   then Content.active
-    when "upcoming" then Content.upcoming
-    when "expired"  then Content.expired
-    when "mine"     then Content.where(user_id: current_user.id)
-    end
-
-    if @query.present?
-      contents_scope = if @scope == "mine"
-        # Mine includes unapproved content which the search corpus excludes,
-        # so fall back to a direct name match against the user's own uploads.
-        like = "%#{ActiveRecord::Base.sanitize_sql_like(@query)}%"
-        contents_scope.where("name LIKE ? COLLATE NOCASE", like)
-      else
-        contents_scope.where(id: Search.matching_ids(@query, Content))
-      end
-    end
+    contents_scope = base_content_scope
+    contents_scope = narrow_by_query(contents_scope) if @query.present?
 
     @contents = policy_scope(contents_scope).includes(:submissions)
   end
@@ -35,5 +20,27 @@ class ContentsController < ApplicationController
   def new
     @content = Content.new
     authorize @content
+  end
+
+  private
+
+  def base_content_scope
+    case @scope
+    when "upcoming" then Content.upcoming
+    when "expired"  then Content.expired
+    when "mine"     then current_user.contents
+    else                 Content.active
+    end
+  end
+
+  # Mine surfaces pending/rejected/unsubmitted uploads, which the search
+  # corpus excludes — so fall back to a direct name match on the user's
+  # own content for that scope.
+  def narrow_by_query(scope)
+    if @scope == "mine"
+      scope.with_name_matching(@query)
+    else
+      scope.where(id: Search.matching_ids(@query, Content))
+    end
   end
 end
--- a/app/controllers/contents_controller.rb
+++ b/app/controllers/contents_controller.rb
@@ -5,23 +5,30 @@ class ContentsController < ApplicationController
 
   # GET /contents or /contents.json
   def index
-    @scope = params[:scope] || "active"
+    @scope = params[:scope].presence_in(%w[active upcoming expired mine]) || "active"
+    # "mine" is meaningless without an authenticated owner; fall back to active.
+    @scope = "active" if @scope == "mine" && !user_signed_in?
     @query = params[:q].to_s.strip
 
     contents_scope = case @scope
-    when "active"
-                  Content.active
-    when "upcoming"
-                  Content.upcoming
-    when "expired"
-                  Content.expired
-    else
-                  Content.active
+    when "active"   then Content.active
+    when "upcoming" then Content.upcoming
+    when "expired"  then Content.expired
+    when "mine"     then Content.where(user_id: current_user.id)
     end
 
-    contents_scope = contents_scope.where(id: Search.matching_ids(@query, Content)) if @query.present?
+    if @query.present?
+      contents_scope = if @scope == "mine"
+        # Mine includes unapproved content which the search corpus excludes,
+        # so fall back to a direct name match against the user's own uploads.
+        like = "%#{ActiveRecord::Base.sanitize_sql_like(@query)}%"
+        contents_scope.where("name LIKE ? COLLATE NOCASE", like)
+      else
+        contents_scope.where(id: Search.matching_ids(@query, Content))
+      end
+    end
 
-    @contents = policy_scope(contents_scope)
+    @contents = policy_scope(contents_scope).includes(:submissions)
   end
 
   # GET /contents/new
--- a/app/views/contents/_grid.html.erb
+++ b/app/views/contents/_grid.html.erb
@@ -8,10 +8,15 @@
           </div>
           
           <div class="space-y-2">
-            <h3 class="text-h4 text-neutral-900 group-hover:text-brand transition-colors">
-              <%= content.name.presence || "Untitled Content" %>
-            </h3>
-            
+            <div class="flex items-start justify-between gap-2">
+              <h3 class="text-h4 text-neutral-900 group-hover:text-brand transition-colors">
+                <%= content.name.presence || "Untitled Content" %>
+              </h3>
+              <% if (badge = moderation_state_badge(content)) %>
+                <%= badge %>
+              <% end %>
+            </div>
+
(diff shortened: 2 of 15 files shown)


app/controllers/contents_controller.rb
```diff
@@ -5,23 +5,30 @@ class ContentsController < ApplicationController
...
-                  Content.active
+    when "active"   then Content.active
+    when "upcoming" then Content.upcoming
+    when "expired"  then Content.expired
+    when "mine"     then Content.where(user_id: current_user.id)
     end
 
-    contents_scope = contents_scope.where(id: Search.matching_ids(@query, Content)) if @query.present?
+    if @query.present?
+      contents_scope = if @scope == "mine"
```
bamnet: This syntax looks pretty weird to me, assigning the result of a multi-line if block to a variable.  Can we clean this up a bit so it's more readable?

app/views/contents/_grid.html.erb
```diff
@@ -8,10 +8,15 @@
...
           <div class="space-y-2">
-            <h3 class="text-h4 text-neutral-900 group-hover:text-brand transition-colors">
-              <%= content.name.presence || "Untitled Content" %>
-            </h3>
-            
+            <div class="flex items-start justify-between gap-2">
+              <h3 class="text-h4 text-neutral-900 group-hover:text-brand transition-colors">
+                <%= content.name.presence || "Untitled Content" %>
+              </h3>
+              <% if (badge = moderation_state_badge(content)) %>
```
bamnet: I don't love this syntax either. Is there a more readable way to express this?

app/views/feeds/show.html.erb
```diff
@@ -123,7 +135,13 @@
           <%= render partial: "contents/grid", locals: {contents: scoped_content} %>
         <% end %>
       <% else %>
-        <% submissions = @feed.submissions.includes(content: :user).where(moderation_status: @status) %>
+        <% submissions = if @status == "mine"
```
bamnet: Again with this weird syntax of assigning variables from multi-line if blocks. I find this difficult to read, can we do something clearer?

app/views/shared/_sidebar.html.erb
```diff
@@ -29,6 +29,9 @@
         <%= sidebar_nav_link_to "Dashboard", "/", "home" %>
         <%= sidebar_nav_link_to "Add Content", new_content_path, "plus" %>
         <%= sidebar_nav_link_to "Browse Content", contents_path, "collection" %>
+        <% if user_signed_in? %>
+          <%= sidebar_nav_link_to "My Content", contents_path(scope: "mine"), "document-text" %>
```
bamnet: I don't actually like this link - it feels duplicative of "Browse Content" above. Let's remove it.

app/controllers/contents_controller.rb
```diff
@@ -5,23 +5,30 @@ class ContentsController < ApplicationController
...
+        # Mine includes unapproved content which the search corpus excludes,
+        # so fall back to a direct name match against the user's own uploads.
+        like = "%#{ActiveRecord::Base.sanitize_sql_like(@query)}%"
+        contents_scope.where("name LIKE ? COLLATE NOCASE", like)
```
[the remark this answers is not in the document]
bamnet: +1. We should find a more Rails-y way to express this.

bamnet:
### Code review

Found 1 issue:

1. N+1 regression: `contents/_grid` now calls `moderation_state_badge(content)` → `Content#moderation_state` → `submissions.map`, but two other callers of the grid partial don't eager-load `:submissions`. `app/views/feeds/show.html.erb` renders the grid with `scoped_content = @feed.approved_content.active/upcoming/expired` (no `.includes`), and `app/views/users/show.html.erb` renders it with `@user.contents`. Each card on those pages will now fire an extra `SELECT submissions WHERE content_id = ?`. Prior PR #1642 had a high-priority callout for exactly this kind of N+1 on the feed show page (https://github.com/concerto/concerto/pull/1642#discussion_r2778402959). `ContentsController#index` was updated with `.includes(:submissions)` in this PR — the other two paths need the same.

https://github.com/concerto/concerto/blob/f765e80f979c289b4eefb3dcb70e0b80e74c3f7f/app/views/contents/_grid.html.erb#L13-L18
https://github.com/concerto/concerto/blob/f765e80f979c289b4eefb3dcb70e0b80e74c3f7f/app/views/feeds/show.html.erb#L114-L136
https://github.com/concerto/concerto/blob/f765e80f979c289b4eefb3dcb70e0b80e74c3f7f/app/views/users/show.html.erb#L90-L94

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

app/controllers/contents_controller.rb
bamnet: Refactored — pulled the search-narrowing into a private `narrow_by_query` and the case statement into `base_content_scope`. The index reads top-to-bottom now.

https://github.com/concerto/concerto/blob/111e57426a42a86ebb96e024066a923dcb1944dd/app/controllers/contents_controller.rb#L7-L41

app/views/contents/_grid.html.erb
bamnet: Dropped the inline-assignment-in-condition. The helper returns nil for approved content, which Rails renders as empty, so `<%= moderation_state_badge(content) %>` works.

https://github.com/concerto/concerto/blob/111e57426a42a86ebb96e024066a923dcb1944dd/app/views/contents/_grid.html.erb#L10-L20

app/views/feeds/show.html.erb
bamnet: Moved the query into `FeedsController#submissions_for_status` so the view just iterates `@submissions`. Same change also eager-loads `content: :user` to keep the submission cards N+1-free.

https://github.com/concerto/concerto/blob/111e57426a42a86ebb96e024066a923dcb1944dd/app/controllers/feeds_controller.rb#L97-L106

app/controllers/contents_controller.rb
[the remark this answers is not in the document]
bamnet: Switched to `LOWER(name) LIKE ?` and wrapped it in a `Content.with_name_matching` scope so the controller call site reads as plain Rails (`scope.with_name_matching(@query)`).

https://github.com/concerto/concerto/blob/111e57426a42a86ebb96e024066a923dcb1944dd/app/models/content.rb#L29-L32

app/views/shared/_sidebar.html.erb
bamnet: Removed. The Mine tab on /contents is the entry point now.

https://github.com/concerto/concerto/blob/111e57426a42a86ebb96e024066a923dcb1944dd/app/views/shared/_sidebar.html.erb#L29-L36

app/controllers/contents_controller.rb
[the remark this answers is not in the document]
bamnet: Done — the controller now calls `scope.with_name_matching(@query)`, which lives on Content alongside the other named scopes.

https://github.com/concerto/concerto/blob/111e57426a42a86ebb96e024066a923dcb1944dd/app/controllers/contents_controller.rb#L29-L41

bamnet:
Pushed [111e5742](https://github.com/concerto/concerto/pull/1813/commits/111e57426a42a86ebb96e024066a923dcb1944dd) addressing all six comments:

- **Readability** (3x): `ContentsController#index` extracts `base_content_scope` + `narrow_by_query`; `_grid.html.erb` uses `<%= moderation_state_badge(content) %>` (nil renders empty); `feeds/show.html.erb` reads `@submissions` set by a new `FeedsController#submissions_for_status` helper.
- **Portability + Rails-y LIKE**: `Content.with_name_matching` scope replaces the inline `COLLATE NOCASE` (now `LOWER(name) LIKE ?`).
- **Sidebar duplication**: My Content link removed; Mine tab on `/contents` remains the entry point.
- **N+1 (from my own review)**: `@feed.approved_content.includes(:submissions)` on the feed approved grid and `@user.contents.includes(:submissions)` on the user profile grid.

990 tests / 2252 assertions pass; rubocop + eslint clean.

PR #1910, opened 2026-07-10
Dedupe content to its best-fit field across a screen (phase 2 of #1906)

### LINKED ISSUE #1906: Content repeated in multiple positions on screen
When a screen has the same feed subscribed to multiple fields, the same content can be rendered in duplicate fields. When there's not much content in the feed, this can result in the same piece of content being duplicated in multiple positions at the same time. This looks really bad.

Example @ https://demo.concerto-signage.org/frontend/2:

- Screen A is using the standard Blue Swoosh template with 4 fields
- Feed A is subscribed to 3 of those fields, everything except for Time.
- Feed A has a mix of large graphic (8.5 x 11 flyer) content suitable for the Main area and text content suitable for the Ticker or Sidebar,

Expected:
- Large graphics are rendered only in the Main area.
- Text content is rendered in either the Ticker or Sidebar, depending on it's size.

Reality:
- Graphics are rendered in both the Main area and the Sidebar.
- Long pieces of text are rendered in all 3 positions, sometimes all at the same time.

bamnet:
## What & why

Phase 2 (and the completing fix) for #1906 — content repeated in multiple positions on a screen. Builds on the `fit_score` groundwork from #1907.

Each field on a screen is rendered by an **independent, stateless request** to `Frontend::ContentController`. Nothing coordinated across fields, so when one feed was subscribed to (say) Main + Sidebar + Ticker, any content that fit all three rendered in all three at once — sometimes the same graphic or blurb in several positions simultaneously.

## Approach

Use the graded `fit_score` to assign each piece of content to the **single field where it fits best**, and render it only there:

- For the content in the current field, gather every field on the screen whose subscribed feed carries it (one `Submission … joins(feed: :subscriptions)` query).
- Score each candidate field via its position in the screen's template.
- The highest-scoring field wins; ties break by lowest field id.
- Keep the content in this field only if this field is the winner. Content that fits nowhere (positive score in no field) is dropped.

Because the winner is a **pure function of the screen's subscriptions + template**, every field's independent request computes the same answer — so content deterministically lands in exactly one field with no shared state or cross-request coordination.

Comparison is done by **field** (not position identity), which is robust to a request's `position_id` differing from the template's position for that field, and correct for the normal 1:1 field↔position case.

## Behavior change

`fetch_subscription_content` no longer gates on `should_render_in?(@position)` directly; the positive-fit requirement is now folded into the best-field selection (a field is only eligible if the content has a positive `fit_score` there). Net effect for a single-field subscription is unchanged; the new behavior only kicks in when a feed spans multiple fields.

## Testing

- New controller tests: content with a feed spanning Main + Sidebar dedupes to the better-fitting field (and each piece lands in exactly one); oversized content subscribed only to small fields is dropped everywhere while a well-sized piece still renders in its best field.
- All existing controller tests unchanged and passing.
- Full suite green (`bin/rails test`: 1075 runs, 0 failures), rubocop clean.

Closes #1906.

### DIFF
--- a/app/controllers/frontend/content_controller.rb
+++ b/app/controllers/frontend/content_controller.rb
@@ -36,16 +36,63 @@ class Frontend::ContentController < Frontend::ApplicationController
   def fetch_subscription_content
     subscriptions = @screen.subscriptions.where(field_id: @field.id)
 
-    # Build content items with subscription metadata and filter by position compatibility
-    content_items = subscriptions.flat_map do |subscription|
-      subscription.contents.active.filter_map do |content|
-        { content: content, subscription: subscription } if content.should_render_in?(@position)
+    items = subscriptions.flat_map do |subscription|
+      subscription.contents.active.map do |content|
+        { content: content, subscription: subscription }
       end
     end
+
+    # Dedupe across fields: when a feed is subscribed to several fields, keep
+    # each piece of content only in the field where it fits best so it isn't
+    # rendered in multiple positions at once. Content that fits nowhere on the
+    # screen is dropped.
+    best_field = best_field_by_content(items.map { |item| item[:content] })
+    content_items = items.select { |item| best_field[item[:content].id] == @field.id }
 
     # Apply ordering strategy (defaults to "random" if no config or blank strategy)
     strategy = @field_config&.ordering_strategy.presence || "random"
     orderer = ContentOrderers.for(strategy)
     orderer.call(content_items)
+  end
+
+  # Maps each content id to the field id on this screen where it fits best.
+  #
+  # A content's candidate fields are those whose subscribed feed carries it.
+  # Each field is scored via its position in the screen's template, and the
+  # highest-scoring field wins (ties broken by lowest field id). Because this
+  # only depends on the screen's subscriptions and template, every field's
+  # independent request computes the same winner, so content lands in exactly
+  # one field. Fields with no template position, or where the content has no
+  # positive fit, are not eligible.
+  def best_field_by_content(contents)
+    content_ids = contents.map(&:id).uniq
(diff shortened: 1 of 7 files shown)


app/controllers/frontend/content_controller.rb
```diff
@@ -36,16 +36,63 @@ def fetch_pinned_content
   def fetch_subscription_content
     subscriptions = @screen.subscriptions.where(field_id: @field.id)
 
-    # Build content items with subscription metadata and filter by position compatibility
-    content_items = subscriptions.flat_map do |subscription|
-      subscription.contents.active.filter_map do |content|
-        { content: content, subscription: subscription } if content.should_render_in?(@position)
```
bamnet, line 42: since we're no longer using `should_render_in?`, is that now dead code we can cleanup?

bamnet:
Observation: In my tests, a 162 character string is rendered in the Main position of standard Blue Swoosh. I'd expect this to land in the sidebar, since it's long but not that long.

app/controllers/frontend/content_controller.rb
bamnet: Good catch to pull on this — it turned out to be more than dead code. `Video` overrode `should_render_in?` directly and never got a `fit_score` in phase 1, so this PR's dedup path (which scores via `fit_score`) was silently falling back to the base `1.0` for videos and bypassing their aspect-ratio filtering. Fixed in 03fd5042 by giving `Video` a graded `fit_score` (mirroring `Graphic`, keeping the 4x tolerance). With Video converted, `should_render_in?` is now unreferenced in `app/`, so I removed it from `Content` and moved the model tests to assert on `fit_score(...).positive?` directly.

PR #1917, opened 2026-07-20
Surface OIDC provisioning failures instead of failing silently

bamnet:
## Problem

A user integrating OIDC reported: *"we are now authenticating with OIDC without error however it appears that it is not logged in."*

Tracing the callback, the root cause is a **silent provisioning failure**:

- `User.from_omniauth` used `first_or_create` (non-bang), which returns an **unpersisted** record when validation fails — no exception.
- The callback treated "not persisted" as "new user needs to register" and redirected to the registration page. No error surfaced anywhere → *authenticated but not logged in*.

Digging in, the actual trigger is narrower than "missing claims" generally: OmniAuth's `InfoHash#name` is **derived** (it falls back to the email address when no name claim is sent), so a missing name rarely blocks provisioning. The real trigger is a **missing `email` claim** — exactly what a bare OIDC deployment does when it only releases a subject/username.

## Changes

- **`omniauth_callbacks_controller.rb`** — on provisioning failure, redirect to sign-in with a clear flash naming the missing claims (and pointing at the `openid`/`email`/`profile` scopes), plus a detailed `[OIDC]` warning log capturing `uid`, `missing_claims`, `received_claims` (what the IdP actually sent), and validation errors — enough to debug without IdP access.
- **`user.rb`** — extracted claim mapping into `assign_name_from_omniauth`; added `missing_omniauth_claims`, which reads the **raw** `name` claim rather than OmniAuth's email-derived one so the diagnostic reflects what the provider truly released.
- **`devise.en.yml`** — two user-facing messages.
- **Tests** — 12 new model tests plus a new controller test file (OmniAuth test mode) covering success, the missing-email failure message, and the debug logging. These paths previously had **no** coverage.

## Behavior note

If a provider sends an email but no name, we still **let the user in** with a name derived from the email rather than blocking. Flagging in case we'd prefer to require a real name instead.

## Testing

`bin/ci` passes: Rails (1090 runs), system (61 runs), seeds, frontend (147 tests), rubocop.


bamnet:
Thanks for the review. Addressed in 9205b28e:

**Critical — blank `provider`/`uid` auth bypass: accepted and fixed.** Confirmed this is real: password-registered users have `NULL` provider/uid, so a blank `uid` made `where(provider: nil, uid: nil).first_or_create` resolve to an existing local account (typically the first system admin, per our bootstrap logic) and sign the caller in as them. `from_omniauth` now rejects a blank/`nil` provider or uid up front and returns an unpersisted, errored record so the callback shows the failure path. Added regression tests covering blank uid (must not match an existing user), blank provider, and nil auth.

**Medium — `NoMethodError` on nil `auth`/`info`: partially applied.** In the real flow these can't be nil — OmniAuth always hands the callback an `AuthHash` whose `.info` auto-vivifies to an (empty) `InfoHash`, and the `openid_connect` action only runs on a *successful* callback — so a crash there would be a 500, not a security issue. That said, now that `from_omniauth` explicitly tolerates malformed input, I made the two pure model helpers (`missing_omniauth_claims`, `assign_name_from_omniauth`) nil/blank-safe, since that's the right home for the robustness. I skipped the controller safe-navigation, as that path isn't reachable with a nil `auth`.

PR #2003, opened 2026-09-17
Block private/internal targets in RemoteFeed's HTTP fetches

### LINKED ISSUE #2002: Server-Side Request Forgery via Remote Feed URL and Remote-Feed-Supplied Image URLs
Affected versions: present in tag v3.0.0.rc3 (commit bcaabb2, 2026-09-01), the latest commit available at time of report.

## Summary

`RemoteFeed#refresh` fetches the feed's configured `url`, and `download_and_attach_image` fetches a per-item image `url` taken from that feed's own JSON response, both via `Net::HTTP` with only an http/https scheme check, no restriction on the destination host. Editing an existing feed and triggering a refresh only requires group-member access (`GroupManagedPolicy#can_edit?`), a lower bar than the group-admin access required to create a new feed. Any such group member can therefore make the Concerto server issue arbitrary HTTP requests to any host reachable from the server, including internal-only services and cloud metadata endpoints.

## Details

`app/models/remote_feed.rb`:

```ruby
validates :url, presence: true, format: { with: URI::DEFAULT_PARSER.make_regexp(%w[http https]) }
...
def refresh
  uri = URI.parse(url)
  http = Net::HTTP.new(uri.host, uri.port)
  ...
end

def download_and_attach_image(graphic, image_url)
  uri = URI.parse(image_url)
  http = Net::HTTP.new(uri.host, uri.port)
  ...
end
```

Nothing rejects loopback, link-local, or private (RFC1918) addresses. `RemoteFeedPolicy` already documents awareness that a feed's URL "may contain sensitive information" and restricts who can view it, but doesn't restrict what it (or an image URL a feed's own response supplies) is allowed to point at.

## POC

(available upon request)

## Impact

A user with group-member (not group-admin) access to a group that owns a Remote Feed can make the server issue arbitrary HTTP requests to internal-only services, cloud provider metadata endpoints, or loopback services on the host itself. Since a feed's own response can additionally supply per-item image URLs fetched via the same unrestricted path, this also enables SSRF via a feed the operator otherwise trusts, independent of the feed's own configured URL.

carfeii:
Fixes #2002

Adds `RemoteFeed.safe_external_uri?`, which resolves the target host and rejects it if any resolved address falls in a private, loopback, link-local, or reserved IP range, and calls it before either HTTP request `RemoteFeed` makes (fetching the feed's own configured URL in `refresh`, and fetching a per-item image URL supplied by the feed's own response in `download_and_attach_image`). Previously neither request path restricted the destination host beyond an http/https scheme check.


bamnet:
(ignore the failing claude code CI, that's on me)

